PR #18, opened 2015-06-21
Add support for multiple Elasticsearch nodes

monicasarbu:
A list of Elasticsearch nodes can be configured by using the `hosts` options. The `host` and `port` options become obsolete, but for backward compatibility, if they are set, a single Elasticsearch node will be used.

  elasticsearch:
    enabled: true
    #host: localhost
    #port: 9200
    hosts: ["localhost:9200", "localhost:9201"]

In case a list of Elasticsearch nodes is configured, for each Elasticsearch operation, a node is selected using the Round Robin technique. If the operation fails, the connection is marked as dead and put on timeout. The timeout is calculated based on the number of failures:

    default_timeout * 2 ** (fail_count - 1) 
  
As long as the connection is marked as dead, the connection is not used. When the timeout is over, the connection is marked as alive. If a connection that is previously marked as dead, succeeds, it is marked as alive and the fail_count is set to zero.

If all the connections are marked as dead, a random connection will be used. 

When an operation fails, it is retried up to max_retries (configurable, by default is 3).

 The algorithm is modeled after the Elasticsearch python client: http://elasticsearch-py.readthedocs.org/en/master/connection.html

### DIFF
--- a/outputs/elasticsearch/api.go
+++ b/outputs/elasticsearch/api.go
@@ -12,6 +12,7 @@ import (
 )
 
 type Elasticsearch struct {
+	MaxRetries int
 	connectionPool ConnectionPool
 	client         *http.Client
 }
@@ -49,6 +50,10 @@ func (r QueryResult) String() string {
 	return string(out)
 }
 
+const (
+	default_max_retries = 3
+)
+
 // Create a connection to Elasticsearch
 func NewElasticsearch(urls []string, username string, password string) *Elasticsearch {
 
@@ -58,6 +63,7 @@ func NewElasticsearch(urls []string, username string, password string) *Elastics
 	es := Elasticsearch{
 		connectionPool: connection_pool,
 		client:         &http.Client{},
+		MaxRetries: default_max_retries,
 	}
 	return &es
 }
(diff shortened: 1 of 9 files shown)


outputs/elasticsearch/api.go
```diff
@@ -104,62 +102,116 @@ func MakePath(index string, doc_type string, id string) (string, error) {
...
+			continue
+		}
+		return resp, nil
+
 	}
 
-	return resp, nil
+	logp.Warn("Request fails to be send after %d retries", es.MaxRetries)
+
+	return nil, fmt.Errorf("Request fails to be sent after %d retries", es.MaxRetries)
```
tsg: We should maybe print what the actual error was here. I know there are multiple errors, I don't know if we should print them all or only the last one.

outputs/elasticsearch/api.go
```diff
@@ -104,62 +102,116 @@ func MakePath(index string, doc_type string, id string) (string, error) {
...
 	if err != nil {
-		return nil, err
+		// request fails
+		es.connectionPool.MarkDead(conn)
+		return nil, fmt.Errorf("Sending the request fails: %s", err)
 	}
 
-	if resp.StatusCode > 299 {
-		return resp, fmt.Errorf("ES returned an error: %s", resp.Status)
+	if resp.StatusCode > 499 {
```
tsg: I noticed that the python client only retries on 503 and 504, and raises error (?) on other 5xx. We should consider doing the same.

PR #21, opened 2015-07-17
Adding support for publishing to Kafka

rshriram:
This patch adds support for publishing log entries directly to a Kafka bus, instead of publishing via redis+logstash. The code uses the sarama Go client library for interfacing with Kafka (shopify/sarama). I would appreciate any feedback you folks may have.

### DIFF
--- a/outputs/kafka/kafka.go
+++ b/outputs/kafka/kafka.go
@@ -1,0 +1,184 @@
+package kafka
+
+import (
+	"encoding/json"
+	"errors"
+	"strings"
+	"time"
+//	"log"
+//	"os"
+
+	"github.com/elastic/libbeat/common"
+	"github.com/elastic/libbeat/logp"
+	"github.com/elastic/libbeat/outputs"
+
+	"github.com/Shopify/sarama"
+)
+
+type KafkaDataType uint16
+
+const (
+	KafkaListType KafkaDataType = iota
+	KafkaChannelType
+)
+
+type KafkaOutput struct {
+	ReconnectInterval  time.Duration
+	BrokerList         []string
+	Topic              string
+	Timeout            time.Duration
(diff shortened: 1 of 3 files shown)


rshriram:
I signed the CLA, btw.

kimchy:
hey @rshriram, thanks for the pull request!. Our plan moving forward is to keep Beats as lightweight as possible. We are going to implement and improve on the lumberjack protocol to Logstash creating a native lightweight integration between the beats and Logstash. Once we do, Redis will not be needed anymore and will be dropped, and the focus will be on lumberjack.

johann8384:
@rshriram Thanks for making this! We pulled it into our Libbeat fork. I'm not saying go use that but I did want to make a note here and let you know that this is useful to other people. We are reading everything from Kafka to write to Elasticsearch, HBase, HDFS and Druid and Logstash isn't actually in our pipeline. So our ELK stack is more like E{LibBeat/Kafka/Storm}K.

davidmoravek:
Can you please reconsider this? Support for kafka is really important, because its log is replayable and can be read by other consumers than logstash (storm, spark, samza, whatever you can imagine). Even if I'd use it only with logstash, it would let me easily reindex all the data just by replaying the message log.

kimchy:
@davidmoravek you can still send the data to logstash which will forward it to Kafka (or any of the many outputs logstash supports). Note, the redis output is deprecated and going to be removed, the way to go is through LS for this using the native LS outputs.

puug:
Does this (seemingly) needlessly complicate the architecture though? In order to use Kafka, Redis, or any other unsupported output, and to keep my app instances lightweight (ie logstash not running locally, only beats) I have to have a HA logstash cluster which beats pushes to, which then forwards to the desired output?

PR #26, opened 2015-07-24
Added a windows service control interface.

### LINKED ISSUE #22: Windows service support
The goal is to run any Beat as a Windows service. For this we need to conform to the interface and requirements of the Service Control Management. 

We would also need the `logp` package to support windows event log.

tsg:
Required for running as a service in Windows. Only responds to the
stop/halt command, nothing else. Closes #22.

### DIFF
--- a/service/service_unix.go
+++ b/service/service_unix.go
@@ -1,0 +1,7 @@
+// +build !windows
+
+package service
+
+// On non-windows platforms, this function does nothing.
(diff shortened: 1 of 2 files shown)


jordansissel:
Nice! I'd historically avoided windows service integration because it felt complicated, but this code isn't scary! I'll see if we can do similar for Logstash soon.

PR #28, opened 2015-07-30
RFC: eliminate snake case and simplify unambiguous names

shutej:
Hi,

I was reading this package and I noticed a few things:

1. There are instances of snake-case in several places.  This is generally not idiomatic in Go.  If camel case is required for reflection mangling, generally people explicitly call it out in struct tags.
2. A few of the identifiers could be simplified, using the fact that Go generally qualifies imports to disambiguate what they are.  For instance, if there's only one interface in a package, you can just call it `Interface`.

As such, I wanted to send this over for comments.  I realize these would be backward-incompatible changes but I could certainly go on and fix packetbeat and other notable consumers (or perhaps write a `go vet` or some such to clean up examples in folks' private repos).

I also notice that in several places the acronyms are capitalized (`IP`) vs. camelcased (`Ip`), but before fixing those wanted to get some comments.

Jeremy

### DIFF
--- a/common/bytes.go
+++ b/common/bytes.go
@@ -8,21 +8,21 @@ import (
 
 // Byte order utilities
 
-func Bytes_Ntohs(b []byte) uint16 {
+func Ntohs(b []byte) uint16 {
 	return uint16(b[0])<<8 | uint16(b[1])
 }
 
-func Bytes_Ntohl(b []byte) uint32 {
+func Ntohl(b []byte) uint32 {
 	return uint32(b[0])<<24 | uint32(b[1])<<16 |
 		uint32(b[2])<<8 | uint32(b[3])
 }
 
-func Bytes_Htohl(b []byte) uint32 {
+func Htonl(b []byte) uint32 {
 	return uint32(b[3])<<24 | uint32(b[2])<<16 |
 		uint32(b[1])<<8 | uint32(b[0])
 }
 
-func Bytes_Ntohll(b []byte) uint64 {
+func Ntohll(b []byte) uint64 {
 	return uint64(b[0])<<56 | uint64(b[1])<<48 |
 		uint64(b[2])<<40 | uint64(b[3])<<32 |
 		uint64(b[4])<<24 | uint64(b[5])<<16 |
@@ -30,7 +30,7 @@ func Bytes_Ntohll(b []byte) uint64 {
 }
 
 // Ipv4_Ntoa transforms an IP4 address in it's dotted notation
-func Ipv4_Ntoa(ip uint32) string {
+func Ipv4Ntoa(ip uint32) string {
 	return fmt.Sprintf("%d.%d.%d.%d",
 		byte(ip>>24), byte(ip>>16),
 		byte(ip>>8), byte(ip))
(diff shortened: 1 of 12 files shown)


shutej:
Oh, one other thing, slices are nil-able types by default, rarely do you need a pointer to a slice.  (Maybe you do in the case I "fixed" but I haven't looked too closely.  Didn't seem like you did, and tests pass.)

common/geolite.go
```diff
@@ -9,27 +9,29 @@ import (
 	"github.com/nranchev/go-libGeoIP"
 )
 
-type Geoip struct {
-	Paths *[]string
```
tsg: This is the pointer to slice instance you mentioned, right? It was there for separating between "paths" being specified in the configuration file but empty and not being present in the configuration file at all. Not sure if there's a better way to do it.

common/bytes.go
```diff
@@ -8,21 +8,21 @@ import (
 
 // Byte order utilities
 
-func Bytes_Ntohs(b []byte) uint16 {
+func Ntohs(b []byte) uint16 {
```
tsg: These functions are used from Packetbeat so we'd need to update that as well.

outputs/outputs.go
```diff
@@ -6,34 +6,34 @@ import (
...
-	Filename           string
-	Rotate_every_kb    int
-	Number_of_files    int
-	DataType           string
-	Flush_interval     *int
-	Bulk_size          *int
-	Max_retries        *int
+type Config struct {
+	Enabled           bool
+	SaveTopology      bool
```
tsg: Here we need json keys to keep the old behavior, otherwise we break the compatibility with the configuration file.

tsg: Actually not json keys, but yaml keys probably.

tsg:
Thanks for the pull request. I'm +1 on these changes but we have to be careful with keeping the compatibility for the configuration file and test the changes with Packetbeat.

common/geolite.go
shutej: So there's a distinction between `g.Paths = nil` and `g.Paths = []string{}`.  The former will be true in a configuration file where you `omitempty`.

All this stuff should really be covered by a unit test, I think.  That's the only way I can be relatively sure that this parses the config the way it should.  Maybe there's an example config in another repo you could point to?

common/bytes.go
shutej: Are they used in other places, yet?  If so I could do a `go fix` (messed that up and typed `go vet` before, different beast).

outputs/outputs.go
shutej: If you point to an example config our two I'll test this change still works the same.  That's likely to yield the best results.

urso:
I'm a big fan of cleaning this up, but 1. other beats MUST be fixed too (use gorename + PR for all affected beats) and 2. config file parsing must not break (having #127 should be done before).

While getting the naming right, we might want to have a look at go metalinter output too.

tsg:
These changes are happening but step wise in different PRs. As we are planning to merge the beats repositories into a single one this week, I'm closing this one because I don't want it to be implicitly lost.

PR #31, opened 2015-08-03
Added paragraph in the Contributing guide about new outputs

tsg:
Add a few words to avoid people putting work in new output plugins.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,11 @@ know about before implementing the change.
 
 We enjoy working with contributors to get their code accepted. There are many
 approaches to fixing a problem and it is important to find the best approach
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -14,7 +14,11 @@ know about before implementing the change.
 
 We enjoy working with contributors to get their code accepted. There are many
 approaches to fixing a problem and it is important to find the best approach
-before writing too much code.
+before writing too much code. In particular, we are very likely to reject pull
```
jordansissel: +1 to this.

If users push back on this in the future, we can work to create better documentation for this use case to show users more easily how to do things with beats/logstash integration.

PR #33, opened 2015-08-13
Speed up pool timeout tests

agonzalezro:
Setting the dead timeout of the pool to 0 the waiting time required for
the test is going to pass from 10 second to 1 millisecond.

I am sure there was a reason for the 10 seconds timeout. If it is still
needed I would recommend at least to check for `testing.Short` on this
test case.


agonzalezro:
I have signed the CLA with this email and this github username. I don't know why the CLA check is failing.

andrewkroh:
@agonzalezro  Thanks. I rebased your PR and merged it in.

PR #34, opened 2015-08-15
Docker Development Environment

ruflin:
DO NOT MERGE

The test suite of libbeat depends on redis and elasticsearch. The goal was that everyone can start to develop and test the full project with as few commands as possible. To simplify the full test suite I introduced docker-compose and Docker environment to setup and run the test suite with just one command (assuming docker is installed):

```
make testsuite
````

This command fetches the images, runs the images and all tests. I suggest that this should be the command that every contributor runs before he opens a pull request as it guarantees that all his changes work in a "standardised" environment.

## Environment
The tests were refactored to fetch the hosts and ports from the environment variables. This was necessary for the dynamic docker environment. In addition, it allows to use the docker redis and elasticsearch image also for the local development. This can be done with the following commands:

```
export ES_HOST=IP_OF_YOUR_DOCKER_MACHINE
export REDIS_HOST=IP_OF_YOUR_DOCKER_MACHINE
make run-environment
```
Like this the local golang environment can be used but still all tests work without any setup of redis or elasticsearch.

### Environment Shell
As an addition for mac users, I added a make command for a dev shell: ```dev-shell```. This only works locally and on Mac's as far as I know. This shares the local volume with the docker image, means all changes on both sides. Like this the image does not have to be rebuilt every time.

## No Gopath required
An interesting side affect of the setup above is that as long as all commands are run inside the virtual environment, the repository can be placed anywhere and doesn't have to be in the go path.

## To be improved
Currently with every changes, part of the images have to be rebuilt as it checks for the dependencies which could have changed. This could be optimized to have an even faster build of the image.

## Documentation
It was not sure where to put the documentation for the above. I suggest to put it in a second step into CONTRIBUTING.md


ruflin:
Closing in favor of https://github.com/elastic/libbeat/pull/54

PR #35, opened 2015-08-17
console output

urso:
Add stdout output type to pretty print published json objects to console. Mostly useful for debugging + testing without having an elasticsearch/redis instance ready.


tsg:
I usually do this with `packetbeat -e -d "publish"` which pretty prints the published json's to stdout. So not sure we want to add a new output plugin for debugging only. Might be interesting if we want have the beats chainable (e.g. packetbeat | catbeat), but not sure there's any use case served by that.

urso:
Thanks, haven't been aware of -d "publish". Like it even better for debugging.

Yes, other use case is to make it chainable (e.g. https://stedolan.github.io/jq/). For example allow users to push all data into their own scripts doing some more magic on transactions.

PR #38, opened 2015-08-21
Feature/protocol helpers1

urso:
Move some helpers for memcached support from packetbeat to libbeat.

Please do not merge this pull request until tests have been add to these modules.

Contains:
- applayer: provides some structures with common fields used by application layer protocols + initialisation of common fields + setting common fields in MapStr for publishing

- streambuf: parse buffer for appending tcp/udp payloads to + some simple re-entrant helper routines for parsing a continues stream of data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ _cgo_export.*
 _testmain.go
 
 *.exe
+
+*.test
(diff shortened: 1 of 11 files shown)


common/applayer/applayer.go
```diff
@@ -0,0 +1,123 @@
+package applayer
```
urso: Put into it's own module to break import cycles with streambuf.
Plus applayer.Stream is shorter than common.AppLayerStream.

tsg: +1

PR #40, opened 2015-08-28
Use the beat name as default index in Elasticsearch

monicasarbu:
Pass the beat name to the publisher init function in order to be used as default
index if none is set in the configuration file.


tsg:
LGTM. @ruflin is working on a more general change regarding this, but I think we should merge this to solve the immediate issue for topbeat.

PR #41, opened 2015-09-01
Introduce MakeDebug

urso:
Add MakeDebug to logp module. Use MakeDebug to create dedicated per module debug functions.

Instead of writing:

    logp.Debug("mymodule", "log message 1")
    ....
    logp.Debug("mymodule", "log message 2")

use MakeDebug to create on per module logging function still printing correct line numbers:

    var debug = logp.MakeDebug("mymodule")
    ....
    debug("log message 1")
    ....
    debug("log message 2")

### DIFF
--- a/logp/log.go
+++ b/logp/log.go
@@ -37,7 +37,7 @@ type Logger struct {
 
 var _log Logger
 
-func Debug(selector string, format string, v ...interface{}) {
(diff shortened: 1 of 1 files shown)


urso:
@tsg @ruflin please review

PR #42, opened 2015-09-01
Add beat abstraction

ruflin:
Add the beat abstraction layer so it can be used by packetbeat and topbeat.

As soon as this is merged, the file beat/beat.go can be removed from packetbeat and topbeat.


monicasarbu:
+1 merging

PR #43, opened 2015-09-01
Developer guide: create new Beat

tsg:
Explains the overall architecture and goes through the code of Topbeat. Let me know what you think.


jordansissel:
Is there a rendered version of this available anywhere to read? Otherwise I am happy to read the asciidoc. :)

tsg:
Good point :-).

Github renders it somehow [decent](https://github.com/tsg/libbeat/blob/devguide/docs/newbeat.asciidoc). Otherwise, it's pretty easy to setup the docs building env. Clone https://github.com/elastic/docs and do something like: 

    ../../docs/build_docs.pl --doc newbeat.asciidoc --single --open

docs/newbeat.asciidoc
```diff
@@ -0,0 +1,287 @@
+= Developer guide: creating a new Beat
+
+This guide walks you through the steps for creating a new Elastic Beat.
+The Beats are a collection of lightweight daemons that collection
```
urso: s/that collection/for collecting/

ruflin: that collect

docs/newbeat.asciidoc
```diff
@@ -0,0 +1,287 @@
...
+collects the actual data and the Publisher. They communicate via a Go channel,
+over which *events* are sent. The Publisher is implemented in libbeat, so you
+typically only have to worry about the logic specific to your Beat. Besides the
+Publisher, libbeat offers common services like configuration management,
+logging, daemonizing/windows service handling and in the future it will offer
+data processing modules like filtering or sampling.
+
+image:./images/beat_overview.png[Beat overview architecture]
+
+An event is JSON like object (Golang type `map[string]interface{}`) that
```
ruflin: is a JSON

docs/newbeat.asciidoc
```diff
@@ -0,0 +1,287 @@
...
+The `Setup` method gives you the opportunity to execute things before the main
+loop, usually for initialization. In the case of Topbeat, it only links the
+`events` channel with the publisher queue. This is the event channel mentioned
+in the Overview section.
+
+[source,go]
+----------------------------------------------------------------------
+func (tb *Topbeat) Setup(b *beat.Beat) error {
+
+	tb.events = publisher.Publisher.Queue
```
ruflin: @tsg @monicasarbu Perhaps we should move this step even to beat.go as the channel is always needed. Just realised that when I was reading the docs.

docs/newbeat.asciidoc
```diff
@@ -0,0 +1,287 @@
...
+Inside the loop, Topbeat sleeps for a configured period of time and then
+captures the required data and sends it to the publisher via the `events`
+channel.
+
+The actual sending is done inside the `exportSystemStats()` function, here is an
+example code snippet:
+
+[source,go]
+----------------------------------------------------------------------
+  event := common.MapStr{
```
ruflin: I suggest to include also the function definition as part of the code. Like this it gets obvious it is related to the comment above.

docs/newbeat.asciidoc
tsg: +1, it was initially in Config, I moved it to Setup (see PR for topbeat). Makes more sense to me to have that in the `beat` package.

ruflin:
Looks good to me. I think it has all the basic information inside. One thing I would add is mentioning that every Beat needs a unique name and version.

Not sure if we should also mention that we recommend to put config parts into config.go

docs/newbeat.asciidoc
ruflin: I will take care of it. https://github.com/elastic/libbeat/issues/44

tsg:
Thanks for the review, I addressed the issues.

ruflin:
I suggest we wait for the review of @jordansissel . Ok?

tsg:
Ugh, actually I was hoping to have this one merged faster so that I can test it together with the rest of the documentation. I think the docs build system doesn't allow using things from a PR, and testing with it is pretty time consuming.

I'm happy to still receive comments on it, but I'd like to have something merged so I can continue with the next step.

ruflin:
Merged :)

tsg:
Thanks. @jordansissel, hope you don't mind!

PR #45, opened 2015-09-02
CodeCov Support

ruflin:
* Align .travis.yml with topbbeat and packetbeat
* Fix a bug that was discovered by `go vet`


cfgfile/cfgfile.go
```diff
@@ -23,7 +23,7 @@ func Read(out interface{}) error {
 		return fmt.Errorf("Fail to read %s: %v. Exiting.", *configfile, err)
 	}
 	if err = yaml.Unmarshal(filecontent, out); err != nil {
-		fmt.Errorf("YAML config parsing failed on %s: %v. Exiting.", *configfile, err)
+		return fmt.Errorf("YAML config parsing failed on %s: %v. Exiting.", *configfile, err)
```
tsg: Ooooh. I think that might explain a problem I had on windows and couldn't figure it out.

PR #48, opened 2015-09-02
Fix handle signals.

tsg:
It needs to be called before the Run function, otherwise it is
essentially never reached. This was especially a problem on Windows
were the service is checked via "signals" that is responsive, and
killed otherwise.

### DIFF
--- a/beat/beat.go
+++ b/beat/beat.go
@@ -3,13 +3,14 @@ package beat
 import (
 	"flag"
 	"fmt"
+	"os"
+	"runtime"
+
 	"github.com/elastic/libbeat/cfgfile"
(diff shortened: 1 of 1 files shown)


beat/beat.go
```diff
@@ -103,11 +104,12 @@ func (b *Beat) Run() {
 	}
 	service.BeforeRun()
 
+	// Callback is called if the processes is asked to stop
+	service.HandleSignals(b.BT.Stop)
```
ruflin: Can you add a comment here with the note that to Windows that you mentioned in the issue description. It was here in the beginning and I moved it down as I wasn't aware about the side affects of changing the order.

Issue #44, opened 2015-09-02
Move events publisher from *beats to libbeat


ruflin:
As the publisher event is used in all beats, it should be part of libbeat. See discussion here: https://github.com/elastic/libbeat/pull/43/files#diff-4d9120a08ab6029b712b996780b4eb4cR179

* Implement it in libbeat and then update topbeat and packetbeat accordingly.
* Apply fixes to libbeat: https://github.com/elastic/topbeat/pull/16/files

ruflin:
Closing as https://github.com/elastic/libbeat/pull/51 is merged.

PR #51, opened 2015-09-03
Event Channel as part of Beat

ruflin:
The documentation will be updated in a second step as soon as topbeat and packetbeat are updated.

See https://github.com/elastic/libbeat/issues/44

### DIFF
--- a/beat/beat.go
+++ b/beat/beat.go
@@ -48,7 +48,6 @@ func NewBeat(name string, version string, bt Beater) *Beat {
 		BT:      bt,
 	}
 
-	b.Events = publisher.Publisher.Queue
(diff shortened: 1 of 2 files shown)


ruflin:
There seems to be an issue with this change. Please don't merge yet.

urso:
Instead of passing a channel how about using an interface having a publish method. Makes it easier to swap the publisher in unit tests for example (have unit test publisher which just collects all events in an array)

ruflin:
@urso I agree we should refactor this. I fixed it in the above commit by just assigning it after the init phase.

urso:
+1

PR #53, opened 2015-09-03
Dependency Management

ruflin:
Add Dependency Management to Contribution file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,3 +44,33 @@ Start in a new repository and use libbeat packages as you would use any other
 Go library. Have a look at the way
 [Packetbeat](https://github.com/elastic/packetbeat) makes use of these packages
 for an example.
+
+
+## Dependency Management
+
+Beats are using [godeps](https://github.com/tools/godep) for dependency management. 
+This means all dependencies are part of the repository. For updating dependencies we
+have the following strategy:
+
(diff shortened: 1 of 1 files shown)


urso:
godep save ./... can be usefull if you for example changed an import path of another project (renamed or use other fork).

Maybe add some not about using 'godep restore' to get the repositorries dependencies into $GOPATH.

If you want to use 'godep save ./...' or 'godep update ./...', make sure 'godep restore' has been run before.

tsg:
I'd say we should try to keep this text small, I don't think we can document godep better than the readme file in the godep repo. Provide a simple workflow and even if someone messes it up, we can fix it relatively easy.

urso:
ok

CONTRIBUTING.md
```diff
@@ -44,3 +44,33 @@ Start in a new repository and use libbeat packages as you would use any other
...
+using 
+
+`go get your-go-package-path` 
+
+or by having the package already in the `$GOPATH`with the correct version / tag.
+To then save the most recent packages into Godep, run 
+
+`godep update your-go-package-path`
+
+Never use `godep save ./...` or `godep update ...` as this will update all packages at
```
tsg: I'd change "Never use" with a softer "Avoid using" because one of these commands might be the right solution for a particular situation.

PR #54, opened 2015-09-04
Environment to run clean test suite with one command

ruflin:
The test suite of libbeat depends on redis and elasticsearch. The goal was that everyone can start to develop and test the full project with as few commands as possible. To simplify the full test suite I introduced docker-compose and Docker environment to setup and run the test suite with just one command (assuming docker is installed):

```
make testsuite
````

This command fetches the images, runs the images and all tests. I suggest that this should be the command that every contributor runs before he opens a pull request as it guarantees that all his changes work in a "standardised" environment. It makes it possible to run everything also on a remote docker-machine.

## Environment
The tests were refactored to fetch the hosts and ports from the environment variables. This was necessary for the dynamic docker environment. In addition, it allows to use the docker redis and elasticsearch image also for the local development. This can be done with the following commands:

```
export ES_HOST=IP_OF_YOUR_DOCKER_MACHINE
export REDIS_HOST=IP_OF_YOUR_DOCKER_MACHINE
make run-environment
```
Like this the local golang environment can be used but still all tests work without any setup of redis or elasticsearch.

The current setup is with Golang 1.4 and the latest release version of elasticsearch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 GODEP=$(GOPATH)/bin/godep
 
 .PHONY: build
-build: 
+build:
 	go get github.com/tools/godep
-	#$(GODEP) go build
+	$(GODEP) go build ./...
 
 .PHONY: deps
 deps:
@@ -40,10 +40,34 @@ cover:
 	go get github.com/pierrre/gotestcover
 	GOPATH=$(shell $(GODEP) path):$(GOPATH) $(GOPATH)/bin/gotestcover -coverprofile=profile.cov -covermode=count github.com/elastic/libbeat/...
 	mkdir -p cover
-	$(GODEP)  go tool cover -html=profile.cov -o cover/coverage.html
+	$(GODEP) go tool cover -html=profile.cov -o cover/coverage.html
 
 .PHONY: clean
 clean:
 	make gofmt
 	-rm profile.cov
 	-rm -r cover
+
+
+# Builds the environment to test libbeat
+.PHONY: build-image
+build-image:
+	make clean
+	docker-compose build
+
+# Runs the environment so the redis and elasticsearch can also be used for local development
+# To use it for running the test, set ES_HOST and REDIS_HOST environment variable to the ip of your docker-machine.
+.PHONY: run-environment
+run-environment: build-image
+	docker-compose run -d
+
+# Runs the full test suite and puts out the result. This can be run on any docker-machine (local, remote)
+.PHONY: testsuite
+testsuite: build-image
+	docker-compose run libbeat make testlong
+	docker ps -a
+	# Copy coverage file back to host
+	mkdir -p cover
+	docker cp libbeat_libbeat_run_1:/go/src/github.com/elastic/libbeat/profile.cov $(shell pwd)/profile.cov
+	docker cp libbeat_libbeat_run_1:/go/src/github.com/elastic/libbeat/cover/coverage.html $(shell pwd)/cover/coverage.html
+	docker-compose stop
--- a/Makefile
+++ b/Makefile
@@ -60,14 +60,17 @@ build-image:
 .PHONY: run-environment
 run-environment: build-image
 	docker-compose run -d
+	
+.PHONY: stop-environment
+stop-environment:
+	docker-compose stop
 
 # Runs the full test suite and puts out the result. This can be run on any docker-machine (local, remote)
 .PHONY: testsuite
 testsuite: build-image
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -40,10 +40,34 @@ cover:
...
+.PHONY: build-image
+build-image:
+	make clean
+	docker-compose build
+
+# Runs the environment so the redis and elasticsearch can also be used for local development
+# To use it for running the test, set ES_HOST and REDIS_HOST environment variable to the ip of your docker-machine.
+.PHONY: run-environment
+run-environment: build-image
+	docker-compose run -d
```
urso: why no stop-environment?

Makefile
```diff
@@ -40,10 +40,34 @@ cover:
...
+# Runs the full test suite and puts out the result. This can be run on any docker-machine (local, remote)
+.PHONY: testsuite
+testsuite: build-image
+	docker-compose run libbeat make testlong
+	docker ps -a
+	# Copy coverage file back to host
+	mkdir -p cover
+	docker cp libbeat_libbeat_run_1:/go/src/github.com/elastic/libbeat/profile.cov $(shell pwd)/profile.cov
+	docker cp libbeat_libbeat_run_1:/go/src/github.com/elastic/libbeat/cover/coverage.html $(shell pwd)/cover/coverage.html
+	docker-compose stop
```
urso: might this stop my test environment I started with run-environment?

In case tests fail, we should also copy the the run directory.

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+    - elasticsearch
+  environment:
+    - ES_HOST=elasticsearch
+    - ES_PORT=9200
+    - REDIS_HOST=redis
+    - REDIS_PORT=6379
+elasticsearch:
+  image: elasticsearch
+  ports:
+    - "9200:9200"
```
urso: got a conflict with my private ES/Kibana instance... ;)

ruflin: Should we do "special ports like 9202 and 6380 to prevent this conflict?

Makefile
ruflin: I will add it

docker-compose.yml
urso: Just for testing I don't think you have to export the ports as the libbeat image will see elasticsearch instance due to linking.

I would say, keep the redis and elasticsearch server only for testing (mostly clean if possible). Not sure I would like to use them for development. I would even run 'docker-compose rm' after finishing the testsuite to have a clean environment always.

Makefile
```diff
@@ -40,10 +40,34 @@ cover:
...
+# Runs the environment so the redis and elasticsearch can also be used for local development
+# To use it for running the test, set ES_HOST and REDIS_HOST environment variable to the ip of your docker-machine.
+.PHONY: run-environment
+run-environment: build-image
+	docker-compose run -d
+
+# Runs the full test suite and puts out the result. This can be run on any docker-machine (local, remote)
+.PHONY: testsuite
+testsuite: build-image
+	docker-compose run libbeat make testlong
```
ruflin: There are two parts here:
* Service images
* libbeat image

The libbeat image has no volumes and is rebuilt every time the source changes. So it should not be an issue here.

For the service images if the container is not removed, they will have the same state when they were shut down. For example elasticsearch has still the indices from last time inside. At the moment the important part for most tests is to have the services available. But you are right if you want to have it 100% reproducible, the containers would have to be removed after every run. Should we do this? Unfortunately there is no nice "docker-compose" command for this :-1:

Makefile
ruflin: Only one environment can run per composer file. You wan't be able to start 2 environments as they are on the same ports.

I will copy over the run directory after the tests.

Makefile
urso: Aren't 'docker-compose rm' or 'docker-compose run --rm' not ok?

Not sure, but I think running 'docker-compose up' will drop old the old containers getting you some fresh state.

My concern here was more about running:
  $ make start-environment
  $ make testsuite

Since 'make testsuite' will reuse same containers as 'make start-environment' does. But not something critical, as start-environment serves another use-case.

ruflin: I actually added docker-compose rm (see in make stop-environment). This helps.

As now make testsuite always cleans up after itself, the only issue is that it can't be run at the same time.

urso: I don't think it's a real problem, with both serving more or less disjunct use cases.

PR #56, opened 2015-09-07
Lumberjack support

urso:


outputs/redis/redis.go
```diff
@@ -117,7 +136,7 @@ func (out *RedisOutput) Init(beat string, config outputs.MothershipConfig, topol
 	logp.Info("[RedisOutput] Using db %d for storing topology", out.DbTopology)
 	logp.Info("[RedisOutput] Using %d data type", out.DataType)
 
-	out.sendingQueue = make(chan RedisQueueMsg, 1000)
+	out.sendingQueue = make(chan message, 1000)
```
tsg: Not necessarily for this PR, but we should consider making these queues configurable or have them at a lower value (logstash uses queues of size 20).

publisher/publish.go
```diff
@@ -11,10 +11,13 @@ import (
...
-	"github.com/elastic/libbeat/outputs/elasticsearch"
-	"github.com/elastic/libbeat/outputs/fileout"
-	"github.com/elastic/libbeat/outputs/redis"
 	"github.com/nranchev/go-libGeoIP"
+
+	// load supported output plugins
+	_ "github.com/elastic/libbeat/outputs/elasticsearch"
+	_ "github.com/elastic/libbeat/outputs/fileout"
+	_ "github.com/elastic/libbeat/outputs/lumberjack"
+	_ "github.com/elastic/libbeat/outputs/redis"
```
tsg: Nice, I like the use of `init()` for this.

outputs/outputs.go
```diff
@@ -30,43 +31,65 @@ type MothershipConfig struct {
...
 }
 
-func (o OutputPlugin) String() string {
-	if int(o) >= len(OutputNames) {
-		return "impossible"
+func InitOutputs(
+	beat string,
+	configs map[string]MothershipConfig,
+	topology_expire int,
+	fn func(string, MothershipConfig, Outputer) error,
```
tsg: This `fn` callback and the `setTopology` thing seem a bit hacky. Wouldn't it be possible to call to do the topology work after InitOutputs?

tsg:
So far LGTM. Do you plan to add the whole lumberjack code in this PR, or have multiple smaller ones?

outputs/outputs.go
```diff
@@ -30,43 +31,65 @@ type MothershipConfig struct {
...
 	// Get the agent name with a specific IP from the topology map
 	GetNameByIP(ip string) string
+}
 
-	// Publish event
-	PublishEvent(ts time.Time, event common.MapStr) error
+type OutputBuilder interface {
+	// Create and initialize the output plugin
+	NewOutput(
+		beat string,
```
ruflin: We could call this beatString so it is not confused with the beat object

outputs/outputs.go
```diff
@@ -30,43 +31,65 @@ type MothershipConfig struct {
...
+	enabledOutputPlugins[name] = builder
 }
 
-func (o OutputPlugin) String() string {
-	if int(o) >= len(OutputNames) {
-		return "impossible"
+func InitOutputs(
+	beat string,
+	configs map[string]MothershipConfig,
+	topology_expire int,
```
ruflin: I suggest we always use lowerUpper for variable names. I just discovered there is golint :-)

ruflin:
@urso I really like the new implementation for Output with the interfaces. It cleans up a lot and makes adding new interfaces even easier.

outputs/redis/redis.go
urso: Just create enhancement request #57

outputs/outputs.go
urso: Indeed, kinda hacky. But name + matching config are lost. InitOutputs will return []OutputPlugin in future:

    type OutputPlugin struct {
        name string
        config MothershipConfig
        output Outputer
    }

outputs/outputs.go
urso: +1 ... I got confused for a second ;)

outputs/outputs.go
urso: yeah, copy'n paste. Try to keep camel-case all the time.

I've been thinking about adding some linter tool. If golint does a good job, let's use it.

urso:
@tsg I was planning to do the lumberjack plugin in here. But for client/server protocol implementation I've been thinking to create an extra repo (like logstash does: one repo protocol implementation and one with actual plugins). Not sure anymore if this was a good idea. Still initial implementation should be relatively small and become part of this pull request. Enhancements will follow in later PRs

outputs/outputs.go
tsg: Ok, sounds good.

urso:
Add first draft of lumberjack output plugin mostly with similar features to LSF. Some todos:

1. test and fix errors
2. add load balancer support
3. either change Outputer interface to support batches or create BatchOutputer interface option for use with filebeat
4. make max window size configurable (so user can reduce max window - batch - size if timeouts happen often)

I know, PR has become quite big. Still point 1 is required before merging. Everything else can be done in subsequent PRs.

Notable differences to LSF:
- support for non-TLS connections
- support for partial ACK. Unfortunately LSF does not evaluate ACKs, so partial ACKs MUST not implemented in logstash input plugin to be backwards compatible
- slow start (window size = 10) with increasing window sizes + window size shrinking on timeout + up to 3 retries before closing connection
- split transport layer, application layer and connection handling in output plugin for extensibility

ruflin:
@urso I know it is rather late, but it would have been nice to have the Output interfaces in an other pull request. Like this, the lumberjack pull request is smaller and can separate the two in case of issues.

urso:
@ruflin 
> I know it is rather late, but it would have been nice to have the Output interfaces in an other pull request. Like this, the lumberjack pull request is smaller and can separate the two in case of issues.

I can't follow you here. Can you elaborate? Where is the pull in here?

ruflin:
@urso I was talking about the Outputer Interface you introduced and the refactoring done to the other outputs.

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (c *tcpClient) Close() error {
+	err := c.Conn.Close()
+	c.connected = false
+	return err
+}
+
+func loadTLSConfig(config *TLSConfig) (*tlsConfig, error) {
+	var tlsconfig tlsConfig
+	tlsconfig.MinVersion = tls.VersionTLS10
```
andrewkroh: I would rather see this default to TLS 1.1 since a lot of security standards are starting to deprecate TLS 1.0.

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+	// Server configurations must include at least one certificate.
+	Certificates []tls.Certificate
+
+	// RootCAs defines the set of root certificate authorities
+	// that clients use when verifying server certificates.
+	// If RootCAs is nil, TLS uses the host's root CA set.
+	RootCAs *x509.CertPool
+}
+
+var ErrNotACertificate = errors.New("File is no certificate")
```
andrewkroh: Suggested wording change: `File is not a certificate`

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+
+		tlsconfig.RootCAs.AddCert(cert)
+	}
+
+	return &tlsconfig, nil
+}
+
+func newTLSClient(host string, tls tlsConfig) (*tlsClient, error) {
+	submatch := hostportRegex.FindSubmatch([]byte(host))
```
andrewkroh: Use `net.SplitHostPort` instead.

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+	Certificates []tls.Certificate
+
+	// RootCAs defines the set of root certificate authorities
+	// that clients use when verifying server certificates.
+	// If RootCAs is nil, TLS uses the host's root CA set.
+	RootCAs *x509.CertPool
+}
+
+var ErrNotACertificate = errors.New("File is no certificate")
+var ErrInvalidAddress = errors.New("Invaliad address")
```
andrewkroh: `s/Invaliad address/Invalid` or `s/Invaliad address/Invalid host and port/`
But I think this goes away if you use `net.SplitHostPort` to parse the host/port pair.

outputs/lumberjack/transport.go
urso: Kept this at 1.0 cause logstash-forwarder is using 1.0 too? But I guess using with 1.1 is better here.

outputs/lumberjack/transport.go
urso: oh, nice. Thanks

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+		c.Close()
+	}
+
+	var tlsconfig tls.Config
+	tlsconfig.MinVersion = c.tls.MinVersion
+	tlsconfig.RootCAs = c.tls.RootCAs
+	tlsconfig.Certificates = c.tls.Certificates
+	tlsconfig.ServerName = c.host
+
+	// TODO: address lookup copied from logstash-forwarded. Really required?
```
andrewkroh: I would be interested to know why it was done this way. `DialTimeout` handles all of this. 

One difference that I can see is that this code randomizes the address selection, but DNS should be responsible for determining the order in which records are returned and hence used. 

Also this code only attempts to use one of the addresses returned by the name lookup, whereas if `DialTimeout` performed the lookup it would serially (possibly in parallel) attempt to connect to each address.

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+	var addressport string
+	addresses, err := net.LookupHost(c.host)
+	if err != nil {
+		logp.Warn("DNS lookup failure \"%s\": %s", c.host, err)
+		return ErrDNSLookupFailed
```
andrewkroh: Any reason for not just passing the `err` from `LookupHost` straight through?

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	address := addresses[rand.Int()%len(addresses)]
+	ip := net.ParseIP(address)
+	if len(ip) == net.IPv4len {
+		addressport = fmt.Sprintf("%s:%s", address, c.port)
+	} else if len(ip) == net.IPv6len {
+		addressport = fmt.Sprintf("[%s]:%s", address, c.port)
+	}
+
+	tcp, err := net.DialTimeout("tcp", addressport, timeout)
```
andrewkroh: Use `net.JoinHostPort`, it should handle bracketing IPv6 addresses. `net.DialTimeout("tcp", net.JoinHostPort(address, c.port), timeout)`

outputs/lumberjack/transport.go
jordansissel: We hung back at TLS1.0 for some reason I don't entirely remember - something about Logstash 1.4's version of JRuby not supporting TLS1.1 properly, or something. I believe it's fixed in newer versions of logstash, but we can test this to find out.

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+		if err != nil {
+			logp.Critical("Failed loading client certificate", err)
+			return nil, err
+		}
+		tlsconfig.Certificates = []tls.Certificate{cert}
+	}
+
+	if len(config.CA) > 0 {
+		tlsconfig.RootCAs = x509.NewCertPool()
+		pemdata, err := ioutil.ReadFile(config.CA)
```
andrewkroh: Consider `s/pemdata/pemData/`.

outputs/lumberjack/transport.go
urso: we should rather test first. Will keep it on 1.0 then until we are sure 1.1 works. Give we want to add a new data frame, no need to be backwards compatible.

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+	"regexp"
+	"time"
+
+	"github.com/elastic/libbeat/logp"
+)
+
+type TLSConfig struct {
+	Certificate string
+	Key         string
+	CA          string
```
andrewkroh: `CA` should be an array. I think it is pretty rare to have only one certificate in your signing chain, unless of course you are signing them with your own CA.

urso: I guess that's the original use case: sign with your own CA

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+	IsConnected() bool
+}
+
+type tcpClient struct {
+	host      string
+	connected bool
+	net.Conn
+}
+
+type tlsClient struct {
```
andrewkroh: Could `tlsClient` just embed `tcpClient`?

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+	tlsconfig.MinVersion = tls.VersionTLS10
+
+	if len(config.Certificate) > 0 && len(config.Key) > 0 {
+		cert, err := tls.LoadX509KeyPair(config.Certificate, config.Key)
+		if err != nil {
+			logp.Critical("Failed loading client certificate", err)
+			return nil, err
+		}
+		tlsconfig.Certificates = []tls.Certificate{cert}
+	}
```
andrewkroh: I am thinking this should return an error if Certificate is set but Key is not, or vice versa. Thoughts?

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,203 @@
...
+			return nil, err
+		}
+
+		block, _ := pem.Decode(pemdata)
+		if block == nil {
+			logp.Critical("Failed to decode PEM. Is certificate %s valid?", config.CA)
+			return nil, ErrNotACertificate
+		}
+		if block.Type != "CERTIFICATE" {
+			logp.Critical("PEM File %s is not a certificate", config.CA)
```
andrewkroh: Consider `s/File/file/`

outputs/outputs.go
```diff
@@ -29,6 +29,10 @@ type MothershipConfig struct {
 	Flush_interval     *int
 	Bulk_size          *int
 	Max_retries        *int
+	Tls                *bool
```
andrewkroh: After looking at transport.go I think this should be `TLS` rather than `Tls`. I struggle to accept this part of the go conventions.

outputs/lumberjack/transport.go
jordansissel: The reason I added randomization on records was to try and use randomness (combined with ack timeouts) to spread load. Without this, you're guaranteed to get a thundering herd on startup because of the way many DNS servers (caching or otherwise) behave. Specific problem here was that at my previous job, I had thousands of logstash-forwarders talking to a fleet of 7 logstash servers, and to avoid an upgrade/update causing a thundering herd to a single server, I implemented random selection on both the servers list as well as in dns (I think; though a former coworker added dns randomization after a discussion we had together). 

(in the event that there is only 1 server entry but that dns name resolves to, for example, several A records)

There are a few RFCs that encourage and discuss reasons for ignoring the order of records returned by the DNS server:

*  http://www.ietf.org/rfc/rfc3484.txt is specific to locality-based address selection for multiple dns records, but exists as prior art in "Choose some different ordering for the dns records" done by a client
* https://tools.ietf.org/html/rfc1794 "Clients, of course, may reorder this information" - with respect to handling order of dns records in a response.

I was unaware until just now that DialTimeout supported parallel and serial modes of connection - this appears undocumented, but is obvious if I look at the Go source code for the net package.

Still, I don't know if we want DialTimeout's parallel/serial mode anyway - I beleive the desired behavior is that if the current connection failed, we probably should try a _different_ server, and DialTimeout could hand us a new connection to the currently-failing server again - my code also could do this, but we could control the behavior, where we cannot with DialTimeout.

My memory is fuzzy on some of this, but the original C version of logstash forwarder had similar behavior (dns lookups, hit a random server): https://github.com/elastic/logstash-forwarder/blob/769bf914f30faccfca84d678b977341d468f35a9/proto.c#L191-L205

Hope this helps fill in some of the history.

I do think we'll still want to do a lookup ourselves and choose a server to talk to for right now, though Stefan and I have discussed future improvements that probably make this not an issue.

andrewkroh: @jordansissel  Thanks. That helps a lot!

outputs/lumberjack/modes.go
```diff
@@ -0,0 +1,189 @@
+package lumberjack
+
+// modes.go implements different sending modes for any kind of ProtocolClient.
+// Currently we support the fail-over like model used by logstash-forwarded only
+//
+// TODO: load-balancing mode with support for scheduling messages by copmuted
```
andrewkroh: s/copmuted/computed/

outputs/lumberjack/modes.go
```diff
@@ -0,0 +1,189 @@
+package lumberjack
+
+// modes.go implements different sending modes for any kind of ProtocolClient.
+// Currently we support the fail-over like model used by logstash-forwarded only
```
andrewkroh: Move the `only`, like "Currently we _only_ support". And might as well drop a `.` on the end.

outputs/lumberjack/transport.go
urso: done

outputs/lumberjack/transport.go
urso: they did differ a little, but tcpClient embedded in tlsClient now

outputs/lumberjack/transport.go
```diff
@@ -0,0 +1,190 @@
...
+		return err
+	}
+
+	var tlsconfig tls.Config
+	tlsconfig.MinVersion = c.tls.MinVersion
+	tlsconfig.RootCAs = c.tls.RootCAs
+	tlsconfig.Certificates = c.tls.Certificates
+	tlsconfig.ServerName = host
+
+	// TODO: address lookup copied from logstash-forwarded. Really required?
```
andrewkroh: I think we have an answer on this. Care to summarize Jordan's feedback?

outputs/lumberjack/lumberjack.go
```diff
@@ -26,16 +29,77 @@ func (p lumberjackOutputPlugin) NewOutput(
...
+	if config.TLS != nil {
+		useTls = *config.TLS
+	}
+
+	timeout := lumberjackDefaultTimeout
+	if config.Timeout != 0 {
+		timeout = time.Duration(config.Timeout) * time.Second
+	}
+
+	var clients []ProtocolClient = nil
```
andrewkroh: You can drop the ` = nil` since that is the zero value of a slice.

outputs/lumberjack/lumberjack.go
```diff
@@ -26,16 +29,77 @@ func (p lumberjackOutputPlugin) NewOutput(
...
+const lumberjackDefaultTimeout = 5 * time.Second
+
+var waitRetry = time.Duration(1) * time.Second
 
 func (lj *lumberjack) init(
 	beat string,
 	config outputs.MothershipConfig,
 	topology_expire int,
 ) error {
+	useTls := false
```
andrewkroh: Consider `s/useTls/useTLS/`.

outputs/lumberjack/lumberjack.go
```diff
@@ -26,16 +29,77 @@ func (p lumberjackOutputPlugin) NewOutput(
...
+	mode ConnectionMode
+}
+
+const lumberjackDefaultTimeout = 5 * time.Second
+
+var waitRetry = time.Duration(1) * time.Second
 
 func (lj *lumberjack) init(
 	beat string,
 	config outputs.MothershipConfig,
```
andrewkroh: This method doesn't check to see if `config.Host` was used rather than `config.Hosts`. Is that on purpose?

outputs/lumberjack/lumberjack.go
```diff
@@ -26,16 +29,77 @@ func (p lumberjackOutputPlugin) NewOutput(
...
+	if useTls {
+		tlsConfig, err := loadTLSConfig(&TLSConfig{
+			Certificate: config.Certificate,
+			Key:         config.CertificateKey,
+			CAs:         config.CAs})
+		if err != nil {
+			return err
+		}
+
+		for _, host := range config.Hosts {
```
andrewkroh: It looks like it is possible for `config.Hosts` to be empty?

outputs/lumberjack/modes.go
```diff
@@ -0,0 +1,189 @@
+package lumberjack
+
+// modes.go implements different sending modes for any kind of ProtocolClient.
+// Currently we onl ysupport the fail-over like model used by logstash-forwarded.
```
andrewkroh: `s/onl ysupport/only support/`

outputs/lumberjack/lumberjack.go
urso: TBH I think Hosts and Host is kinda redundant. The redis outputer only uses Host, whereas elasticsearch outputer uses Host as fallback if Hosts is not set.

I would vote for getting rid of Host and use Hosts only. Either in some kind of load-balancing mode or fail-over mode if connection goes down.

outputs/lumberjack/lumberjack.go
urso: check for len()==0 missing.

outputs/lumberjack/lumberjack.go
andrewkroh: I would favor removing `Host` too. I didn't know it existed until now.

outputs/lumberjack/client.go
```diff
@@ -0,0 +1,282 @@
...
+	minWindowSize             int = 1
+	maxWindowSize             int = 1024
+	defaultStartMaxWindowSize int = 10
+
+	maxAllowedTimeoutErr int = 3
+)
+
+// errors
+var (
+	ErrNoAckResponse = errors.New("Invalid response is no ACK")
```
andrewkroh: So this error indicates that a response was received, but it is not an ACK?

urso: bad name. The protocol would expect an ACK (only message type the server can send right now). But the response received is not an ACK.

outputs/lumberjack/client.go
```diff
@@ -0,0 +1,282 @@
...
+	if count == 0 {
+		// encoding of all events failed. Let's stop here and report all events
+		// as exported so no one tries to send/encode the same events once again
+		// The compress/encode function already prints critical per failed encoding
+		// failure.
+		return len(events), nil
+	}
+
+	// send window size:
+	var ackSeq uint32
```
andrewkroh: I suggest moving `ackSeq` down closer to where it is first used.

outputs/lumberjack/client.go
```diff
@@ -0,0 +1,282 @@
...
+// lumberjackClient implements the ProtocolClient interface to be used
+// with different mode. The client implements slow start with low window sizes +
+// window size backoff in case of long running transactions.
+//
+// it is suggested to use lumberjack in conjunction with roundRobinConnectionMode
+// if logstash becomes unresponsive
+type lumberjackClient struct {
+	TransportClient
+	windowSize      int
+	maxOkWindowSize int
```
andrewkroh: Can you document `maxOkWindowSize`. It wasn't clear to me what its value meant.

outputs/lumberjack/client.go
```diff
@@ -0,0 +1,282 @@
...
+	l.windowSize = l.windowSize / 2
+	if l.windowSize < minWindowSize {
+		l.windowSize = minWindowSize
+	}
+	return int(ackSeq), nil
+}
+
+func (l *lumberjackClient) compressEvents(events []common.MapStr) (uint32, []byte) {
+	buf := bytes.NewBuffer(nil)
+	buf.Truncate(0)
```
andrewkroh: I haven't used this package before. Why does it need to `Truncate` a new Buffer?

urso: Ups. Doesn't need to truncate. I used to have a shared buffer stored within the client before.

outputs/lumberjack/client_test.go
```diff
@@ -0,0 +1,373 @@
...
+
+func sendAck(transp *mockTransport, seq uint32) {
+	buf := streambuf.New(nil)
+	buf.WriteByte('1')
+	buf.WriteByte('A')
+	buf.WriteNetUint32(seq)
+	transp.sendBytes(buf.Bytes())
+}
+
+func Test_SendZero(t *testing.T) {
```
ruflin: A general not: Golang recommends the test function name without the _ https://golang.org/pkg/testing/

ruflin:
@urso Quite nice and clean code here. +1 from me.

outputs/lumberjack/transport.go
urso: will do, but will implement same logic for tcp first.

outputs/lumberjack/client.go
urso: I can not due to 'goto fail'. Will have to refactor error handling first.

andrewkroh: Oh. that makes sense. I definitely didn't think about the `goto` there. I like the refactoring.

outputs/lumberjack/transport.go
urso: @jordansissel what's the minimal supported version of JRuby?

tsg:
LGTM.

andrewkroh:
LGTM.

PR #59, opened 2015-09-09
Refactor/output plugin

urso:


ruflin:
Looks good to me. I first wanted to comment that we should not have lumberjack in with this pull request already, but at the same time it already has the foundation in.

PR #64, opened 2015-09-09
Refactor and test Read config option to support files input

ruflin:
Filebeat requires multiple config files to be read. The current config read implementation picked the path to read directly from the command line input. Now the path is put in through a parameter. Some tests were added to verify the changes.

### DIFF
--- a/cfgfile/cfgfile.go
+++ b/cfgfile/cfgfile.go
@@ -17,13 +17,21 @@ func CmdLineFlags(flags *flag.FlagSet, name string) {
 	testConfig = flags.Bool("test", false, "Test configuration and exit.")
 }
 
-func Read(out interface{}) error {
-	filecontent, err := ioutil.ReadFile(*configfile)
+// Reads config from yaml file into the given interface structure.
+// In case the second param path is not set
+func Read(out interface{}, path string) error {
+
+	if path == "" {
+		path = *configfile
+	}
+
+	filecontent, err := ioutil.ReadFile(path)
+
 	if err != nil {
-		return fmt.Errorf("Fail to read %s: %v. Exiting.", *configfile, err)
+		return fmt.Errorf("Fail to read %s: %v. Exiting.", path, err)
 	}
 	if err = yaml.Unmarshal(filecontent, out); err != nil {
-		return fmt.Errorf("YAML config parsing failed on %s: %v. Exiting.", *configfile, err)
+		return fmt.Errorf("YAML config parsing failed on %s: %v. Exiting.", path, err)
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


cfgfile/cfgfile.go
```diff
@@ -17,13 +17,21 @@ func CmdLineFlags(flags *flag.FlagSet, name string) {
 	testConfig = flags.Bool("test", false, "Test configuration and exit.")
 }
 
-func Read(out interface{}) error {
-	filecontent, err := ioutil.ReadFile(*configfile)
+// Reads config from yaml file into the given interface structure.
+// In case the second param path is not set
```
andrewkroh: Missing the end of this sentence. How about... `In case path is not set this method reads from the default configuration file for the beat.`

cfgfile/cfgfile.go
```diff
@@ -17,13 +17,21 @@ func CmdLineFlags(flags *flag.FlagSet, name string) {
...
+
+	if path == "" {
+		path = *configfile
+	}
+
+	filecontent, err := ioutil.ReadFile(path)
+
 	if err != nil {
-		return fmt.Errorf("Fail to read %s: %v. Exiting.", *configfile, err)
+		return fmt.Errorf("Fail to read %s: %v. Exiting.", path, err)
```
andrewkroh: I think `s/Fail/Failed/` would read better.

tests/files/config.yml
```diff
@@ -0,0 +1,4 @@
+server:
+  elasticsearch:
+    port: 9200
+    host: "localhost"
```
andrewkroh: Are the quotes necessary here?

cfgfile/cfgfile.go
ruflin: Ups, forgot to finish the sentence.

tests/files/config.yml
ruflin: I wanted to make sure it is a string, but let me test if it works also without.

ruflin: Works also without quotes.

ruflin:
Changes applied, lets wait for travis.

cfgfile/cfgfile.go
```diff
@@ -17,13 +17,21 @@ func CmdLineFlags(flags *flag.FlagSet, name string) {
 	testConfig = flags.Bool("test", false, "Test configuration and exit.")
 }
 
-func Read(out interface{}) error {
-	filecontent, err := ioutil.ReadFile(*configfile)
+// Reads config from yaml file into the given interface structure.
```
andrewkroh: Consider starting the method doc with the method name like described in [google’s conventions](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences). We might want to decide as a team if we want to follow these conventions. Feel free to ignore.

andrewkroh:
LGTM

PR #65, opened 2015-09-10
Make all command line flags use the global flag Flagset

ruflin:
The global flag object is used for the flag options as otherwise the flags are not discovered by the integration tests. 

This means packetbeat and topbeat also have to be updated. Pull requests are here:
* https://github.com/elastic/packetbeat/pull/237
* https://github.com/elastic/topbeat/pull/32

### DIFF
--- a/cfgfile/cfgfile.go
+++ b/cfgfile/cfgfile.go
@@ -12,9 +12,10 @@ import (
 var configfile *string
 var testConfig *bool
 
-func CmdLineFlags(flags *flag.FlagSet, name string) {
-	configfile = flags.String("c", fmt.Sprintf("/etc/%s/%s.yml", name, name), "Configuration file")
-	testConfig = flags.Bool("test", false, "Test configuration and exit.")
+func init() {
+	// The default config cannot include the beat name as it is not initialised when this function is called
+	configfile = flag.String("c", "/etc/beat/beat.yml", "Configuration file")
+	testConfig = flag.Bool("test", false, "Test configuration and exit.")
 }
 
 // Read reads the configuration from a yaml file into the given interface structure.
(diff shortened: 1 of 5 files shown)


cfgfile/cfgfile.go
```diff
@@ -12,9 +12,10 @@ import (
 var configfile *string
 var testConfig *bool
 
-func CmdLineFlags(flags *flag.FlagSet, name string) {
-	configfile = flags.String("c", fmt.Sprintf("/etc/%s/%s.yml", name, name), "Configuration file")
-	testConfig = flags.Bool("test", false, "Test configuration and exit.")
+func init() {
+	// The default config cannot include the beat name as it is not initialised when this function is called
```
andrewkroh: That is unfortunate. This could lead to configuration confusion if you have two different Beats installed.

ruflin: I plan to overwrite this when config and flags are merged. I also opened an issue for that to make sure it doesn't get forgotten: https://github.com/elastic/libbeat/issues/66

andrewkroh: Great, then LGTM.

tsg:
From the commit log, this was the reason for not using the global flagset:

> The global FlagSet from the global flag module can be
>    polluted by libraries. For example, the glog library (imported
>   via go-redis) adds its own flags for verbosity which can
>  confuse our users.

The good news is that we don't use go-redis anymore so that's no longer an issue. Just mentioning it to keep an eye for flags added by other libraries.

ruflin:
@tsg That is actually a point a really worry about, especially if flags conflict. As a partial solution I suggest to use "namespaces" for all beat flags. For example "packetbeat.devices". The part it still doesn't solve is that the user sees flags that are not from us.

@andrewkroh Can you merge it?

PR #68, opened 2015-09-11
Feature/bulk publisher

urso:
Publish a batch of events in one call (only lumberjack implements the bulk interface so far)

depends on #56 due to lumberjack implemention for bulk interface


monicasarbu:
Looks good to me.

PR #69, opened 2015-09-11
Feature/transactional publisher

urso:
Define outputs.Transactioner interface to signal success/failure of output/publish operations. Beats must pass nil if they are not interested in any status output.

All output plugins have been modified to signal transactioners after doing IO. If publisher has multiple output plugins configured and only one interface fails, the complete transaction will be marked as Failed().

depends on #56 and #68.

### DIFF
--- a/outputs/elasticsearch/api.go
+++ b/outputs/elasticsearch/api.go
@@ -59,7 +59,7 @@ const (
 func NewElasticsearch(urls []string, username string, password string) *Elasticsearch {
 
 	var connection_pool ConnectionPool
-	connection_pool.SetConnections(urls, username, password)
+	_ = connection_pool.SetConnections(urls, username, password) // never errors
 
 	es := Elasticsearch{
 		connectionPool: connection_pool,
@@ -175,7 +175,7 @@ func (es *Elasticsearch) PerformRequest(conn *Connection, req *http.Request) ([]
 		return nil, do_retry, fmt.Errorf("%v", resp.Status)
 	}
 
-	defer resp.Body.Close()
+	defer func() { _ = resp.Body.Close() }()
 	obj, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		es.connectionPool.MarkDead(conn)
--- a/outputs/elasticsearch/api.go
+++ b/outputs/elasticsearch/api.go
@@ -22,7 +22,7 @@ type QueryResult struct {
 	Ok      bool            `json:"ok"`
 	Index   string          `json:"_index"`
 	Type    string          `json:"_type"`
-	Id      string          `json:"_id"`
+	ID      string          `json:"_id"`
 	Source  json.RawMessage `json:"_source"`
 	Version int             `json:"_version"`
 	Found   bool            `json:"found"`
@@ -52,26 +52,26 @@ func (r QueryResult) String() string {
 }
 
 const (
-	default_max_retries = 3
+	defaultMaxRetries = 3
 )
 
-// Create a connection to Elasticsearch
+// NewElasticsearch creates a connection to Elasticsearch.
 func NewElasticsearch(urls []string, username string, password string) *Elasticsearch {
 
-	var connection_pool ConnectionPool
-	_ = connection_pool.SetConnections(urls, username, password) // never errors
+	var pool ConnectionPool
+	_ = pool.SetConnections(urls, username, password) // never errors
 
 	es := Elasticsearch{
-		connectionPool: connection_pool,
+		connectionPool: pool,
 		client:         &http.Client{},
-		MaxRetries:     default_max_retries,
+		MaxRetries:     defaultMaxRetries,
 	}
 	return &es
 }
 
 // Encode parameters in url
-func UrlEncode(params map[string]string) string {
-	var values url.Values = url.Values{}
+func urlEncode(params map[string]string) string {
+	values := url.Values{}
 
 	for key, val := range params {
 		values.Add(key, string(val))
@@ -79,15 +79,15 @@ func UrlEncode(params map[string]string) string {
 	return values.Encode()
 }
 
-// Create path out of index, doc_type and id that is used for querying Elasticsearch
-func MakePath(index string, doc_type string, id string) (string, error) {
+// Create path out of index, docType and id that is used for querying Elasticsearch
+func makePath(index string, docType string, id string) (string, error) {
 
 	var path string
-	if len(doc_type) > 0 {
+	if len(docType) > 0 {
 		if len(id) > 0 {
-			path = fmt.Sprintf("/%s/%s/%s", index, doc_type, id)
+			path = fmt.Sprintf("/%s/%s/%s", index, docType, id)
 		} else {
-			path = fmt.Sprintf("/%s/%s", index, doc_type)
+			path = fmt.Sprintf("/%s/%s", index, docType)
 		}
 	} else {
 		if len(id) > 0 {
@@ -103,7 +103,7 @@ func MakePath(index string, doc_type string, id string) (string, error) {
 	return path, nil
 }
 
-func ReadQueryResult(obj []byte) (*QueryResult, error) {
+func readQueryResult(obj []byte) (*QueryResult, error) {
 
 	var result QueryResult
 	if obj == nil {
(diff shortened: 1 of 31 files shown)


outputs/elasticsearch/output.go
```diff
@@ -285,10 +219,61 @@ func (out *elasticsearchOutput) UpdateLocalTopologyMap() {
...
-func (out *elasticsearchOutput) PublishEvent(ts time.Time, event common.MapStr) error {
-
-	out.sendingQueue <- EventMsg{Ts: ts, Event: event}
+func (out *elasticsearchOutput) PublishEvent(
+	signaler outputs.Signaler,
+	ts time.Time,
+	event common.MapStr,
+) error {
+	index := fmt.Sprintf("%s-%d.%02d.%02d",
+		out.Index, ts.Year(), ts.Month(), ts.Day())
```
andrewkroh: I don't know if this is needed because I haven't tested it yet, but thought I would at least bring it up. Should this convert the timestamp with `ts.UTC()` before calling `Day()`, `Month()`, and `Year()`?

outputs/elasticsearch/api.go
```diff
@@ -155,27 +155,27 @@ func (es *Elasticsearch) PerformRequest(conn *Connection, req *http.Request) ([]
...
 			es.connectionPool.MarkDead(conn)
-			do_retry = true
+			doRetry = true
 		}
-		return nil, do_retry, fmt.Errorf("%v", resp.Status)
+		return nil, doRetry, fmt.Errorf("%v", resp.Status)
 	}
 
-	defer resp.Body.Close()
+	defer func() { _ = resp.Body.Close() }()
```
ruflin: What is the difference to the one before?

outputs/elasticsearch/output.go
urso: This is no new code, but actually copying/refactoring already present code. We should discuss/fix this separately.

outputs/elasticsearch/api.go
urso: no real difference, but makes clear we are ignoring the error.

We could replace this by something like 'defer common.Closing(resp.Body)'.

outputs/elasticsearch/bulkapi.go
```diff
@@ -69,20 +58,126 @@ func (es *Elasticsearch) BulkRequest(method string, path string,
...
 
-	return ReadQueryResult(resp)
+func bulkEncode(metaBuilder MetaBuilder, body []interface{}) bytes.Buffer {
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
+	if metaBuilder == nil {
+		for _, obj := range body {
+			pos := buf.Len()
+			if err := enc.Encode(obj); err != nil {
+				debug("Failed to encode message: %s", err)
```
ruflin: Shouldn't this be logp.Debug(*?

urso: var debug = logp.MakeDebug("publish")

So usage of 'debug(...)' is equivalent to 'logp.Debug("publish", ...)'

After having had some typos I found a per module debug function easier to use.

outputs/elasticsearch/output.go
```diff
@@ -134,7 +120,10 @@ func (out *elasticsearchOutput) Init(beat string, config outputs.MothershipConfi
 func (out *elasticsearchOutput) EnableTTL() error {
 
 	// make sure the .packetbeat-topology index exists
-	out.Conn.CreateIndex(".packetbeat-topology", nil)
+	// Ignore error here, as CreateIndex will error (400 Bad Request) if index
+	// already exists. If index could not be created, next api call to index will
+	// fail anyway.
+	_, _ = out.Conn.CreateIndex(".packetbeat-topology", nil)
```
ruflin: This looks very .packetbeat specific. I don't think it should be like this in libbeat. Perhaps we can use the beat name here?

urso: Yes it is. Some more stuff in publisher is packetbeat specific too. Needs to be changed/factored out in another PR

outputs/elasticsearch/output.go
```diff
@@ -285,10 +219,61 @@ func (out *elasticsearchOutput) UpdateLocalTopologyMap() {
...
+			err := request.Send(meta, event)
+			if err != nil {
+				logp.Err("Fail to encode event: %s", err)
+			}
+		}
+
+		_, err = request.Flush()
+		outputs.Signal(trans, err)
+		if err != nil {
+			logp.Err("Fail to perform many index operations in a single API call: %s",
```
ruflin: Improved error message: "Failed to send elasticsearch bulk request. Error: %s"

urso: +1

outputs/lumberjack/client.go
```diff
@@ -167,11 +176,13 @@ func (l *lumberjackClient) compressEvents(events []common.MapStr) (uint32, []byt
 			sequence-- //forget this last broken event and continue
 		}
 	}
-	compressor.Flush()
-	compressor.Close()
+	if err := compressor.Close(); err != nil {
+		debug("Finalizing zlib compression failed with: %s", err)
```
ruflin: Should be logp.Debug(?

outputs/lumberjack/client.go
```diff
@@ -221,12 +232,15 @@ func (l *lumberjackClient) writeDataFrame(
 
 	jsonEvent, err := json.Marshal(event)
 	if err != nil {
-		logp.Err("Fail to convert the event to JSON: %s", err)
+		debug("Fail to convert the event to JSON: %s", err)
```
ruflin: There are several more debug( below.

outputs/lumberjack/modes.go
```diff
@@ -134,7 +146,7 @@ func newFailOverConnectionMode(
 		timeout:   timeout,
 		waitRetry: waitRetry,
 	}
-	f.Connect(-1)
+	_ = f.Connect(-1) // try to connect, but ignore errors for now
```
ruflin: I would add a // TODO to this comment as I think we should improve that before the release.

outputs/lumberjack/client.go
urso: debug = logp.Debug

outputs/lumberjack/client.go
urso: debug = logp.Debug

outputs/lumberjack/modes.go
urso: it's not a TODO. This is a preliminary connect. The main-loop will try to connect later. But some error/warn output on failure in Connect method would be nice.

publisher/client.go
```diff
@@ -0,0 +1,84 @@
...
+// ChanClient will forward all published events one by one to the given channel
+type ChanClient struct {
+	Channel chan common.MapStr
+}
+
+type client struct {
+	publisher *PublisherType
+}
+
+type nilClient struct{}
```
ruflin: Is this for testing purpose? If yes should it be moved to testing?

urso: yes, for testing (used by packetbeat). But publisher has no testing module at all yet.

If I add NilClient to client_test.go, will it still be visible for packetbeat unit tests?

ruflin:
LGTM

publisher/client.go
urso: +1

moving NilClient to client_test.go works

outputs/elasticsearch/api.go
ruflin: No, looks good to me. I was just curious what the intention was.

outputs/elasticsearch/output.go
ruflin: :+1: Can you open an issue for that so we don't forget it?

publisher/client_test.go
```diff
@@ -0,0 +1,16 @@
+package publisher
```
ruflin: @urso This works as long as you don't use publisher_test as package.

outputs/elasticsearch/output.go
urso: https://github.com/elastic/libbeat/issues/80

outputs/elasticsearch/output.go
urso: https://github.com/elastic/libbeat/issues/81

urso:
push

ruflin:
@urso From my side ready to merge. But you need to merge master in again ...

PR #71, opened 2015-09-12
Feature/transactions

urso:


urso:
Add support for synchronous transaction handling. Depends on #68 and #69 waiting for the publisher signaling ok/failure when sending data.

outputs/fileout/file.go
```diff
@@ -64,18 +64,21 @@ func (out *fileOutput) init(beat string, config outputs.MothershipConfig, topolo
...
+	trans outputs.Transactioner,
+	ts time.Time,
+	event common.MapStr,
+) error {
+	jsonEvent, err := json.Marshal(event)
 	if err != nil {
-		logp.Err("Fail to convert the event to JSON: %s", err)
-		return err
-	}
+		// mark as success so event is not send again.
```
ruflin: s/send/sent

outputs/outputs.go
```diff
@@ -62,12 +70,37 @@ type OutputInterface interface {
 	TopologyOutputer
 }
 
+type Transactioner interface {
+	// Completed is called by publish/output plugin when all events have been
+	// send
```
ruflin: s/send/sent

outputs/fileout/file.go
```diff
@@ -64,18 +64,21 @@ func (out *fileOutput) init(beat string, config outputs.MothershipConfig, topolo
...
+		outputs.CompleteTransaction(trans)
 
-	err = out.rotator.WriteLine(json_event)
-	if err != nil {
+		logp.Err("Fail to convert the event to JSON: %s", err)
```
ruflin: As this is needed for most outputs, I suggest to implement a "marshal" function in for example output.go that also then directly catches the error. Like this it can be used for all outputs.

outputs/redis/redis.go
```diff
@@ -315,15 +326,20 @@ func (out *redisOutput) UpdateLocalTopologyMap(conn redis.Conn) {
...
+func (out *redisOutput) PublishEvent(
+	trans outputs.Transactioner,
+	ts time.Time,
+	event common.MapStr,
+) error {
 
-	json_event, err := json.Marshal(event)
+	jsonEvent, err := json.Marshal(event)
 	if err != nil {
 		logp.Err("Fail to convert the event to JSON: %s", err)
```
ruflin: see comment above about marshalling function in output.go

outputs/elasticsearch/output_test.go
```diff
@@ -118,7 +118,7 @@ func TestOneEvent(t *testing.T) {
 		},
 	})
 
-	err := elasticsearchOutput.PublishEvent(ts, event)
+	err := elasticsearchOutput.PublishEvent(nil, ts, event)
```
ruflin: What does it exactly mean if the outputs.Transactioner is set to nil? It means it is is not disabled?

urso:
removed 'review' label. Too much async functionality in outputers which might lead to deadlocks.

outputs/fileout/file.go
urso: There is some more functionality hidden inside output plugins that needs to be moved...

outputs/elasticsearch/output_test.go
urso: the Transactioner is kind of an asynchronous return value (signal). Setting it to nil means we are not interested in any signal/return value about a job being complete (empty interface).

urso:
obsolete due to overall refactoring required in publisher and output plugins

PR #72, opened 2015-09-14
Added a Cache data structure to libbeat.

andrewkroh:
This pull request adds a cache data structure to libbeat. The Cache will be used to address some concurrency-safety issues in Packetbeat. 

Packetbeat will be updated to store the streams and transactions of most of the protocol plugins in the Cache. This will address the immediate needs for elastic/packetbeat#203 but the final solution will be to refactor Packetbeat to not require locking. 

Also by using this cache, Packetbeat will not need to create a new `time.Timer` for every transaction. Instead, a single goroutine can be used to periodically clean up expired entries in an instance of the Cache.

### DIFF
--- a/common/cache.go
+++ b/common/cache.go
@@ -1,0 +1,229 @@
+package common
+
+import (
+	"sync"
+	"time"
+)
+
+// Key type used in the cache.
+type Key interface{}
+
+// Value type held in the cache. Cannot be nil.
+type Value interface{}
+
+// RemovalListener is the callback function type that can be registered with
+// the cache to receive notification of the removal of expired elements.
+type RemovalListener func(k Key, v Value)
+
+// Clock is the function type used to get the current time.
+type clock func() time.Time
+
+// An element stored in the cache.
+type element struct {
+	expiration time.Time
+	value      Value
+}
+
+// IsExpired returns true if the element is expired (current time is greater
+// than the expiration time).
+func (e *element) IsExpired(now time.Time) bool {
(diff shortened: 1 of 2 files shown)


ruflin:
@andrewkroh Is this cache packetbeat specific or can it be used by any other beat?

common/cache.go
```diff
@@ -0,0 +1,234 @@
...
+// the cache to receive notification of the removal of expired elements.
+type RemovalListener func(k Key, v Value)
+
+// Clock is the function type used to get the current time.
+type clock func() time.Time
+
+// Clock function that returns the current system time.
+var systemClock clock = func() time.Time {
+	return time.Now()
+}
```
urso: var systemClock clock = time.Now  ?

urso:
+1

common/cache.go
andrewkroh: Good point. I updated the pull request to directly use `time.Now`. Thanks.

andrewkroh:
@ruflin The Cache does not have any Packetbeat specific functionality so in theory it could be used by any beat that needed it. But keep in mind that using it does sacrifice type safety and it uses a Mutex for synchronization.

ruflin:
@andrewkroh Looks good to me. Can you merge in master again. There seems to be a conflict.

andrewkroh:
@ruflin Thanks for reviewing. I rebased my changes on top of master. It should be good to go now.

PR #61, opened 2015-09-09
Final docker-compose implementation for jenkins

ruflin:

PR #76, opened 2015-09-15
Enhancing testsuite build target to make it a bit more reliable.

andrewkroh:
###### Changes
* ~~Fixes an issue with libbeat docker container caching the contents of the repository.~~
* The libbeat container now waits for elasticsearch to start before executing tests.
* ~~Exposing redis and elasticsearch ports on the docker-machine for local development purposes.~~
* Added clean-up command to ensure any stopped container is deleted prior to starting to ensure the libbeat container is predictably named which is required in order to copy coverage reports to host.
* Rename make target `cover` to `coverage`.

The Jenkins job for libbeat should be updated to remove the `sleep 10` and add an explicit `make stop-environment`. `make stop-environment` should be called irregardless of whether or not the build succeeds to ensure that the docker containers are stopped and removed.

###### Running Integration Tests inside Docker
This will start the elasticsearch and redis containers and then execute the integration tests inside an container.
```
make testsuite
```

You should stop the Elasticsearch and Redis containers when you are done with them.
```
make stop-environment
```

###### Running Integration Tests Locally
This method is more complicated but enables you to execute the integration tests from your machine rather than inside of a container.
```
# Verify that the active machine is what you expect.
docker-machine active
docker-compose pull
# Set the environment variables that are read by the tests.
export ES_HOST=$(docker-machine ip <name here>)
export ES_PORT=<choose an available port>
export REDIS_HOST=$(docker-machine ip <name here>)
export REDIS_PORT=<choose an available port>
# Start up containers and expose the ports on the docker machine.
docker run -d -p $ES_PORT:9200 --name libbeat_es elasticsearch
docker run -d -p $REDIS_PORT:6379 --name libbeat_redis redis
# Wait for the containers to boot then run integration tests.
make testlong
```

To stop the containers:
```
docker stop libbeat_es libbeat_redis
docker rm libbeat_es libbeat_redis
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,10 +9,20 @@ RUN go get \
 	golang.org/x/tools/cmd/cover \
 	golang.org/x/tools/cmd/vet
 
-WORKDIR /go/src/github.com/elastic/libbeat
-# Setup work environment
-RUN mkdir -p /go/src/github.com/elastic/libbeat
+COPY docker-entrypoint.sh /entrypoint.sh
 
+# Setup work environment
+ENV LIBBEAT_PATH /go/src/github.com/elastic/libbeat
+RUN mkdir -p $LIBBEAT_PATH
+WORKDIR $LIBBEAT_PATH
+
+# Create a copy of the respository inside the container.
 COPY . /go/src/github.com/elastic/libbeat
 
+# It is expected that libbeat from the host is mounted
+# within the container at the WORKDIR location.
+ENTRYPOINT ["/entrypoint.sh"]
+
+# Build libbeat inside of the container so that it is ready
+# for testing.
 RUN make
--- a/Makefile
+++ b/Makefile
@@ -30,25 +30,25 @@ autotest:
 .PHONY: testlong
 testlong:
 	go vet ./...
-	make cover
+	make coverage
 
 .PHONY: benchmark
 benchmark:
 	go test -short -bench=. ./...
 
-.PHONY: cover
-cover:
+.PHONY: coverage
+coverage:
 	# gotestcover is needed to fetch coverage for multiple packages
 	go get github.com/pierrre/gotestcover
 	GOPATH=$(shell $(GODEP) path):$(GOPATH) $(GOPATH)/bin/gotestcover -coverprofile=profile.cov -covermode=count github.com/elastic/libbeat/...
-	mkdir -p cover
-	$(GODEP) go tool cover -html=profile.cov -o cover/coverage.html
+	mkdir -p coverage
+	$(GODEP) go tool cover -html=profile.cov -o coverage/coverage.html
 
 .PHONY: clean
 clean:
 	make gofmt
 	-rm profile.cov
-	-rm -r cover
+	-rm -r coverage
 
 
 # Builds the environment to test libbeat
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -60,23 +60,22 @@ build-image:
...
 .PHONY: testsuite
 testsuite: build-image
 	docker-compose run libbeat make testlong
 	# Copy coverage file back to host
-	mkdir -p cover
-	docker cp libbeat_libbeat_run_1:/go/src/github.com/elastic/libbeat/profile.cov $(shell pwd)/profile.cov
-	docker cp libbeat_libbeat_run_1:/go/src/github.com/elastic/libbeat/cover/coverage.html $(shell pwd)/cover/coverage.html
-	make stop-environment
+	docker cp libbeat_libbeat_run_1:/go/src/github.com/elastic/libbeat/profile.cov $(shell pwd)/
+	docker cp libbeat_libbeat_run_1:/go/src/github.com/elastic/libbeat/cover/coverage.html $(shell pwd)/cover/
```
ruflin: I'm currently renaming this folders from cover to coverage. Can you do it here also? Will also need some adjustments at other places.

docker-compose.yml
```diff
@@ -8,7 +8,13 @@ libbeat:
     - ES_PORT=9200
     - REDIS_HOST=redis
     - REDIS_PORT=6379
+  volumes:
+    - .:/go/src/github.com/elastic/libbeat
```
ruflin: I on purpose copied the content of the repository instead of using volumes. Volumes does not work for example if your docker-machine runs on Google Cloud Engine. In addition it leaded to some flaky test as some stuff was not written, but never fully found out why.

docker-compose.yml
```diff
@@ -8,7 +8,13 @@ libbeat:
...
 elasticsearch:
   image: elasticsearch
+  ports:
+    - "9200:9200"
```
ruflin: We had initially the ports exposed but removed it again as on startup it conflicts with other "elasticsearch" and "redis" instances which are already running on jenkins.

Dockerfile
```diff
@@ -9,10 +9,13 @@ RUN go get \
 	golang.org/x/tools/cmd/cover \
 	golang.org/x/tools/cmd/vet
 
-WORKDIR /go/src/github.com/elastic/libbeat
-# Setup work environment
-RUN mkdir -p /go/src/github.com/elastic/libbeat
+COPY docker-entrypoint.sh /entrypoint.sh
 
-COPY . /go/src/github.com/elastic/libbeat
```
ruflin: See comment below why I used to copy over the content and not volumes.

ruflin:
@andrewkroh See my comments above. I recommend not to go with volumes.

For the following line:
```
export ES_HOST=$(docker-machine ip)
```
Not sure if you mean here "put your docker-machine ip here" or you mean running the command. Running the command doesn't work as the docker-machine ip commands also needs an additional parameter, the name of the machine.

Makefile
andrewkroh: @ruflin Yep, I will make that change. Are you changing the make target's name too or leaving as `cover`?

docker-compose.yml
andrewkroh: @ruflin I wasn't aware of this. I was thinking you might have had a reason for using `COPY`. The problem I was trying to address was that the container image is being cached with the repo contents inside of it.

We can add `--no-cache` to the docker build command but this will cause it to re-download the various go tools being installed on every build (not optimal IMO). But if we separate the `COPY` command into its own Dockerfile then we will have the best of both worlds (a cachable golang build container + non-cached container with libbeat code inside). Would you be OK with a second Dockerfile for this?

docker-compose.yml
andrewkroh: Woops. I forgot about that. We could default these to a non-standard ports and update the tests to read the `ES_PORT` and `REDIS_PORT` env vars. Or we could remove `ports` from docker-compose.yml and just document how to expose the ports from the docker-machine and leave it up to the user to deconflict ports if needed. Any preference?

Dockerfile
andrewkroh: @ruflin  Thanks for pointing this out to me. I will make sure to switch it back to using `COPY`. Using `VOLUME` was a bad idea for multiple reasons.

andrewkroh:
@ruflin  I will update the command to include name of the docker-machine. However that command does work for me. I am only using a single docker-machine and I have version 0.2.0 installed.

Makefile
ruflin: I call it also coverage

docker-compose.yml
ruflin: @andrewkroh I don't really see the issue with caching. As far as I know, every time a single file on your disk changes, the COPY command is rerun (please double check this). The other option would be what you suggest above, a second Dockerfile. But I think it is actually not needed.

docker-compose.yml
ruflin: The ports were initially introduced here: https://github.com/elastic/libbeat/pull/54
And removed here: https://github.com/elastic/libbeat/pull/61

As you see in the comments, we also discussed about the "random" ports. Even though I like to have it as easy as possible, random ports will only make it confusing I think. Because of this I would go with option three, documentation. Keeping in mind, that people don't read, perhaps we can also offer them a command how to open it locally ...

ruflin:
@andrewkroh I assume it works as you only have one docker-machine installed. Try to add an additional one and check if it still works. I'm running on version 0.4.1.

docker-compose.yml
andrewkroh: @ruflin I re-tested this several times this morning with fresh eyes and each time docker correctly detected the changed files on disk. Reverted my change.

Dockerfile
```diff
@@ -9,10 +9,16 @@ RUN go get \
...
 # Setup work environment
-RUN mkdir -p /go/src/github.com/elastic/libbeat
+ENV LIBBEAT_PATH /go/src/github.com/elastic/libbeat
+RUN mkdir -p $LIBBEAT_PATH
+WORKDIR $LIBBEAT_PATH
 
+# Create a copy of the respository inside the container.
 COPY . /go/src/github.com/elastic/libbeat
 
-RUN make
```
ruflin: I had this as part of the Dockerfile as like this the "binary" is part of the image. So in case no files change, it can already be used.

PR #82, opened 2015-09-17
lumberjack: use json encoded data frame for output

urso:


andrewkroh:
Looks good. It's much nicer not having to flatten the JSON.

ruflin:
I especially like that it makes the code simpler.

Issue #79, opened 2015-09-17
Disparity in TLS config for lumberjack and ES output plugin


urso:
ES relies on host configuration: https://www.elastic.co/guide/en/beats/packetbeat/current/_using_https_and_basic_authentication.html

Whereas lumberjack output plugin requires all TLS based configuration in config file.

Update output plugins to both support TLS via host configuration and optionally config file.

urso:
PR #86

monicasarbu:
Could you please update the documentation as well? I am moving this section to libbeat/doc/https.asciidoc in https://github.com/elastic/libbeat/pull/98

monicasarbu:
Available here as well: https://github.com/elastic/packetbeat/pull/261

monicasarbu:
Track the issue https://github.com/elastic/libbeat/issues/100

monicasarbu:
Duplicated

PR #86, opened 2015-09-21
Bug #79 tls config

urso:
Load TLS configuration for elasticsearch and lumberjack from config file.

If config file specifies no root CAs, the hosts root CAs will be used.

Configure certificate and certificatekey for TLS client authentication.


andrewkroh:
LGTM

PR #91, opened 2015-09-22
es outputer: fix timestamp conversion

urso:


ruflin:
@urso Merged. Perhaps we can add a small unit test here later?

PR #99, opened 2015-09-23
Use the Beat name in the default configuration file path

### LINKED ISSUE #66: Make sure default config is named the same as beat name
See also https://github.com/elastic/libbeat/pull/65/files#r39164410

tsg:
Adds a bit of extra complexity but fortunately requires no changes to the Beats, it's a change internal in libbeat.

Should fix #66.

### DIFF
--- a/beat/beat.go
+++ b/beat/beat.go
@@ -63,7 +63,7 @@ func (beat *Beat) CommandLineSetup() {
 	err := cfgfile.ChangeDefaultCfgfileFlag(beat.Name)
 	if err != nil {
 		fmt.Printf("Failed to fix the -c flag: %v\n", err)
-		os.Exit(0)
+		os.Exit(1)
 	}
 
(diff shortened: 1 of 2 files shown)


tsg:
@ruflin I'd like your review on this one. Others are welcome as well, ofc.

monicasarbu:
LGTM

ruflin:
@tsg This will unfortunately not work with our system tests as the flag must be defined either in the init() function or outside the methods.

tsg:
What does "outside the methods" mean more exactly? Global?

ruflin:
Yes. The better definition is it must run before the main() function runs. This is the case for all init() functions.

tsg:
Came up with a better solution, which works for the coverage-enabled system tests binary. One caveat is that this binary now has a different default value when compared with the regular binary, but that doesn't affect our tests.

PR #104, opened 2015-09-23
lumberjack load balancer mode

urso:
Add mode to balance load between multiple lumberjack connections

### DIFF
--- a/outputs/lumberjack/modes.go
+++ b/outputs/lumberjack/modes.go
@@ -9,6 +9,7 @@ package lumberjack
 import (
 	"errors"
 	"math/rand"
+	"sync"
 	"time"
 
 	"github.com/elastic/libbeat/common"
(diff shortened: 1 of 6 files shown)


tsg:
Can we get a short description of the LB algorithm? Like round robin or something else, reconnect policies, resent policies, etc. Would be useful for the future as well.

outputs/lumberjack/modes.go
```diff
@@ -244,3 +263,124 @@ func (f *failOverConnectionMode) PublishEvents(
...
+		msg.attemptsLeft--
+		if msg.attemptsLeft <= 0 {
+			outputs.SignalFailed(msg.signaler)
+			return
+		}
+
+		select {
+		case m.retries <- msg:
+			return
+		case <-time.After(m.timeout):
```
andrewkroh: It probably makes little difference, but I think if you added `case <-done:` it would shutdown faster on `Close()`.

outputs/signal.go
```diff
@@ -13,6 +13,16 @@ type Signaler interface {
 	Failed()
 }
 
+// ChanSignal will send outputer signals on configurable channel.
```
andrewkroh: I like these additions.

andrewkroh:
LGTM. Nice work!

outputs/lumberjack/modes.go
urso: good point. could hang for quite some time.

tsg:
Thanks for adding the comments!

PR #110, opened 2015-09-25
Add version variable for Elasticsearch, Kibana and beats-dashboards.

monicasarbu:
Configure what version of Elasticsearch, Kibana, Logstash (needs to be done) and beats-dashboard to be downloaded and installed.

https://github.com/elastic/packetbeat/issues/235 suggests to use _latest_ for Elasticsearch packages. For Kibana, there is kibana-latest, but from what I can see it doesn't include a binary file.

I think it's better if don't use the _latest_ and choose the version of Elasticsearch and Kibana that work better with each Beat release.

### DIFF
--- a/docs/gettingstarted.asciidoc
+++ b/docs/gettingstarted.asciidoc
@@ -31,32 +31,32 @@ mac for OS X):
 
 deb:
 
-[source,shell]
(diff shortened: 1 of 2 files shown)


andrewkroh:
LGTM.

PR #112, opened 2015-09-25
Testing environment

ruflin:
Testing environment with a Mock Beat to also create integration tests for libbeat.


.travis.yml
```diff
@@ -5,10 +5,20 @@ go:
...
+
 services:
   - redis-server
   - elasticsearch
 
+addons:
+  apt:
+    packages:
+      - python-virtualenv
+      - geoip-database
```
andrewkroh: Not a really a new issue, but I am pretty sure the geoip-database package does not contain the GeoIP city database that libbeat requires to actually provide coordinates. Any queries to this DB return a named place, but not coordinates.

andrewkroh:
LGTM.

.travis.yml
ruflin: Is this dependency even needed to run the tests?

ruflin:
@urso Good points. We should discuss this further. Also if we want to directly apply it to this pull request or refactor it in a second stage.

PR #114, opened 2015-09-25
enhance lumberjack send mode

urso:
- unify retry handling
- configurable number of retries (of 0, output plugins block until send)
- fix possible deadlock in load balancer, if all worker connections start failing at about the same time.


andrewkroh:
LGTM.

PR #115, opened 2015-09-25
add option Sync to PublishEvent(s)

urso:
new option Sync will block until all events have been successfully
published

### DIFF
--- a/publisher/client.go
+++ b/publisher/client.go
@@ -26,6 +26,7 @@ type client struct {
 
 type publishOptions struct {
 	confirm bool
+	sync    bool
 }
 
 // ClientOption allows API users to set additional options when publishing events.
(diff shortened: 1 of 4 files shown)


publisher/client.go
```diff
@@ -37,6 +38,13 @@ func Confirm(options *publishOptions) {
 	options.confirm = true
 }
 
+// Sync option will block the event publisher until an event has beend ACKed by
```
andrewkroh: `s/beend/been/` and `s/send/sent/`

publisher/client.go
```diff
@@ -37,6 +38,13 @@ func Confirm(options *publishOptions) {
...
+// the output plugin. If output plugin signals failure, the event is send again.
```
andrewkroh: "again" makes me think it will retry once. Maybe say something similar to "If the output plugin signals failure, it will retry until it signals success."

andrewkroh:
LGTM.

PR #118, opened 2015-09-30
Feature/output modes

urso:
extra output modes from lumberjack output plugin for reuse


outputs/mode/balance.go
```diff
@@ -0,0 +1,226 @@
...
+
+// PublishEvent forwards the event to some load balancing worker.
+func (m *LoadBalancerMode) PublishEvent(
+	signaler outputs.Signaler,
+	event common.MapStr,
+) error {
+	return m.publishEventsMessage(eventsMessage{
+		attemptsLeft: m.maxAttempts,
+		signaler:     signaler,
+		event:        event,
```
ruflin: Wouldn't it be simpler to internally just use events but only with 1 item inside?

ruflin:
This looks good to me.

A general question: Wouldn't it be simpler to treat PublishEvent as a proxy for PublishEvents that only has one element? I would assume this could simplify the code quite a bit. In case this is something we should move forward, I would suggest to do it in a different pull request.

urso:
Reason is PublishEvents uses BulkOutput API, whereas PublishEvent uses simple per Event output API. E.g. for elasticsearch output plugin API usage would change. Rather do this kind of refactoring in another PR.

ruflin:
@urso Ok, makes sense. It seems like one of the unit tests is failing. Can you have a look?

urso:
@ruflin unit tests stabilized in dcbc125 and f83369b

PR #123, opened 2015-10-05
testsuite go1.5

urso:
update testsuite to use go1.5 too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 *.swn
 libbeat
 libbeat.test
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -9,6 +9,7 @@
 *.swn
 libbeat
 libbeat.test
+coverage/coverage.html
```
ruflin: I suggest we ignore the whole folder: coverage/*

urso: +1

Issue #100, opened 2015-09-23
Update libbeat doc


monicasarbu:
Common documentation for all beats is moved to libbeat (see #98).
Things that need to be updated in the beats common documentation:
   - [x] Update output plugins to both support TLS via host configuration and optionally config file.
   - [x] Add Lumberjack output 
   - [x] Document GeoIP configuration and requirement to use GeoIP City database.
   - [x] Document which cipher suites are not recommend (see commend @jordansissel in https://github.com/elastic/libbeat/pull/124)

monicasarbu:
Would be nice to add a few info about how the agents can load balance the data to Logstash.

urso:
all done.

PR #124, opened 2015-10-05
expose more TLS config options

urso:
Expose all TLS config options used by golang tls clients via config
file.

TLS config parsing requires go1.5 to build (#123). Requirement due to TLS cipher suites added in go1.5.

### DIFF
--- a/outputs/tls.go
+++ b/outputs/tls.go
@@ -19,13 +19,39 @@ var (
 
 	// ErrKeyNoCertificate indicate a configuration error with missing certificate file
 	ErrKeyNoCertificate = errors.New("certificate file not configured")
+
+	// ErrInvalidTLSVersion indicates an unknown tls version string given.
+	ErrInvalidTLSVersion = errors.New("invalid TLS version string")
+
+	// ErrUnknownCipherSuite indicates an unknown tls cipher suite being used
+	ErrUnknownCipherSuite = errors.New("unknown cypher suite")
+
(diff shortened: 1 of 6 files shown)


urso:
Documentation will be done in another PR (see issue #100)

outputs/tls.go
```diff
@@ -68,14 +94,102 @@ func LoadTLSConfig(config MothershipConfig) (*tls.Config, error) {
...
+	case "1.2":
+		return tls.VersionTLS12, nil
+	default:
+		return 0, ErrInvalidTLSVersion
+	}
+}
+
+func parseTLSCipherSuites(names []string) ([]uint16, error) {
+	suites := map[string]uint16{
+		"RSA-RC4-128-SHA":                tls.TLS_RSA_WITH_RC4_128_SHA,
```
jordansissel: RC4 is not recommended; should we disable it?

urso: same goes for AES. I just keep it in for now to be compatible + expose all known (to golang) config options. Easier to maintain list of known codecs vs. list of recommended codecs...

"Recommendations" we should put into doc?

Optionally will we can add warning if not recommended codec is used.

If cipher suite is not configured, golang will use internal default list with recommended order (Have stdlib authors maintain this list).

ruflin: I added a note to https://github.com/elastic/libbeat/issues/100 to document it.

jordansissel: +1

PR #129, opened 2015-10-07
Switch travis to docker-compose setup

ruflin:
This makes the travis setup identical to the jenkins setup. In addition it tests for two different versions of elasticsearch.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,13 +2,17 @@ libbeat:
   build: .
   links:
     - redis
-    - elasticsearch
+    - elasticsearch-172
+    - elasticsearch-200
   environment:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -2,13 +2,17 @@ libbeat:
   build: .
   links:
     - redis
-    - elasticsearch
+    - elasticsearch-172
+    - elasticsearch-200
```
urso: here we always startup both containers. But should be ok.

ruflin: I know. Unfortunately I didn't find an option to "inject" environment variables into the docker-compose file. There are endless discussions and pull request to do it, but didn't happen yet.

PR #136, opened 2015-10-08
Complete config file based on documentation

ruflin:
All the options from the documentation were copied over to the config file.

Later it should be decided which config options should be kept in. As all outputs take MothershipConfig as config option, any of these options could be set for each output.

Tests for each config option should be added to make sure it works as expected.

This is the base for: https://github.com/elastic/packetbeat/issues/292

### DIFF
--- a/etc/libbeat.yml
+++ b/etc/libbeat.yml
@@ -6,7 +6,7 @@
 # You can enable one or multiple outputs by setting enabled option to true.
 output:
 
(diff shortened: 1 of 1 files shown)


urso:
which issue is resolved by this?

ruflin:
I linked the wrong issue. This is the base for: https://github.com/elastic/packetbeat/issues/292

PR #137, opened 2015-10-08
Add elasticsearch-200 as host default for Makefile. 

ruflin:
Empty host didn't work


urso:
duplicate? Same change was required by #128 to fix build

PR #139, opened 2015-10-13
Doc: TLS config

urso:
enhance and document TLS configuration

### DIFF
--- a/etc/libbeat.yml
+++ b/etc/libbeat.yml
@@ -34,15 +34,6 @@ output:
     # Optional HTTP Path
     #path: "/elasticsearch"
 
-    # List of root certificates for HTTPS server verifications
-    #cas: ["/etc/pki/root/ca.pem"]
-
-    # Certificate for TLS client authentication
-    #certificate: "/etc/pki/client/cert.pem"
-
-    # Client Certificate Key
-    #certificatekey: "/etc/pki/client/cert.key"
-
     # The number of times a particular Elasticsearch index operation is attempted. If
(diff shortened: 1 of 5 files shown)


urso:
Update documentation for #100

ruflin:
@urso Looks like this config change is also a BC change. Should be added to CHANGELOG.md. LGMT,

urso:
good point

etc/libbeat.yml
```diff
@@ -65,11 +56,32 @@ output:
...
-    # Controls whether the client verifies server certificates and host name.
-    # If tlsinsecure is set to true, all server host names and certificates will be
-    # accepted. In this mode TLS based connections are susceptible to
-    # man-in-the-middle attacks. Use only for testing.
-    #tlsinsecure
+    # tls configuration
+    #tls:
+      # If disabled is set to true, the tls section will be ignored and the
+      # host its certificate authorities will be used.
+      #disabled: true
```
ruflin: I prefer to have "positiv" configs like "enabled", but that is a discussion we should have in the code base, not the docs.

ruflin: I prefer to have "positiv" configs like "enabled", but that is a discussion we should have in the code base, not the docs.

urso: it's kind of a 'hidden' option to disable tls (similar effect to commenting out the section). Due to false being the default value for booleans I opted for not naming it enabled. We can remove this flag as well.

PR #144, opened 2015-10-14
Cache: optional per element timeout

urso:
Configure per element timeout in cache. If element timeout is <=0, the
cache its default timeout will be used.

required for elastic/packetbeat#300

### DIFF
--- a/common/cache.go
+++ b/common/cache.go
@@ -86,7 +86,16 @@ func newCache(d time.Duration, initialSize int, l RemovalListener, t clock) *Cac
 // PutIfAbsent writes the given key and value to the cache only if the key is
 // absent from the cache. Nil is returned if the key-value pair were written,
 // otherwise the old value is returned.
-func (c *Cache) PutIfAbsent(k Key, v Value, timeout time.Duration) Value {
+func (c *Cache) PutIfAbsent(k Key, v Value) Value {
+	return c.PutIfAbsentWithTimeout(k, v, 0)
+}
+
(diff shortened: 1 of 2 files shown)


tsg:
There was a discussion on Hipchat about this one, with the following options mentioned:

* provide a `cache.DefaultTimeout` constant for the caller to use instead of 0
* use the var args trick to simulate an optional parameter so you can call `cache.Put(k, v)`, or `Put(k, v, common.Timeout(....))`
* add another `PutWithTimeout(k, v, timeout)` function

My opinion is the last option would be the most idiomatic Go thing to do, but I defer to @urso and @andrewkroh  for the decision.

urso:
I will go for option 3.

McStork:
@urso Could it be possible to set timeout to 0 ? For the use case we have we don't need to link queries with responses; and we might not be the only ones in this situation.

urso:
@McStork timeout of 0 should be invalid. All application layer module correlate requests to responses. You use-case of basically just parsing messages is not support by packetbeat, but if it can not correlate a request to a response, it can still publish the messages.

McStork:
Alright. We actually want to log all the incoming packets, but not the reponses. I set a low timeout so it can publish faster to ES. It's just too bad we cannot disable the correlation request <-> response.

urso:
@McStork feel free to create an enhancement request on packetbeat to start some discussion (not everyone will follow this ticket here).

PR #149, opened 2015-10-14
update tls certificate doc

urso:


urso:
I hope the fact the client certificate being optional is more clear by now.

PR #150, opened 2015-10-14
Add direction field to publisher output to indicate inbound/outbound transactions

rshriram:
When ignore_outgoing is set to false, the outputs at the client and server are very similar, except for the response_time field. To clearly distinguish the direction of transaction (inbound or outbound), add a "direction" field to the publishers output.

Note that its very useful to set ignore_outgoing to false, when packetbeat is installed at both client and server, because (a) at the server side, the response_time field indicates the server's processing time for a given request, (b) at the client side, it provides the client observed response time of a transaction -- which includes Network RTT + server processing time. Subtracting client.response_time - server.response_time provides network RTT information.

### DIFF
--- a/publisher/preprocess.go
+++ b/publisher/preprocess.go
@@ -148,6 +148,15 @@ func updateEventAddresses(publisher *PublisherType, event common.MapStr) bool {
 		delete(event, "dst")
 	}
 
+	//get the direction of the transaction: outgoing (as client)/incoming (as server)
+	if (publisher.IsPublisherIP(dst.Ip)) {
+		// outgoing transaction
+		event["direction"] = "out"
+	} else {
+		//incoming transaction
+		event["direction"] = "in"
+	}
+
 	if publisher.IgnoreOutgoing && dstServer != "" &&
 		dstServer != publisher.name {
 		// duplicated transaction -> ignore it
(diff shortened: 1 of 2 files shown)


rshriram:
The build seems to have stopped at logstash-devutils installation stage. I cannot see any other useful info on Travis.

urso:

urso:
@rshriram build should be fixed. try to rebase on master.

rshriram:
@urso thanks. I rebased earlier in the day. I now see more updates. Should I wait until the code is reviewed before I rebase again?

tsg:
Hi @rshriram, I'm +1 for this change but the new fields needs to be added [here](https://github.com/elastic/packetbeat/blob/master/etc/fields.yml), otherwise the Packetbeat tests will fail. Can you open a PR against Packetbeat to add it, please? I think it's best if you write the description for it.

rshriram:
@tsg done. Its PR 317 in elastic/packetbeat (I don't know how to refer to links on other repos).

urso:
related packetbeat PR elastic/packetbeat#317

The functionality in preprocess.go is still very packetbeat specific and might move some time to packetbeat. For now I would say it's a good place to add this functionality.

+1

tsg:
@urso agreed.

@rshriram thanks for the contribution!

PR #153, opened 2015-10-15
fix elasticsearch host configuration address types

### LINKED ISSUE #142: Fail to configure localhost:9200 as ES node
To be backwards compatible, we need to be able to accept IP:PORT in the hosts config option. If the IP is a string then it fails.

The config file:

      output:
          elasticsearch:
          enabled: true
          hosts: ["localhost:9200"]

You can easily check that by adding a unit test for "localhost:9200".

urso:
fixes #142

This PR adds some more possible address types to the unit tests and fixes parsing by retrying on weird parse results by prepending scheme to given address.


urso:

urso:

monicasarbu:
Great, thanks Steffen.

PR #154, opened 2015-10-15
Update tls config options naming

urso:
Use underscore in tls yaml config options to be consistent with other options.


urso:

PR #165, opened 2015-10-19
Documentation changes for debug selectors and cipher suites.

andrewkroh:


docs/command-line.asciidoc
```diff
@@ -18,9 +18,10 @@ Pass the configuration file of your Beat.
 Write CPU profile to a given file. This option is useful for troubleshooting
 the Beat.
 
-==== -d
-Enable debugging for certail selectors. For example `-d publish` displays all
-the publish related messages. 
+==== -d [selectors]
+Enable debugging for the specified selectors. The selectors are a comma-separated
+list of components for which to enable debug. Use `*` to enable debug for all
+components. For example, `-d publish` displays all the "publish" related messages.
```
ruflin: Should we add here a list of available selectors, at least for libbeat?

ruflin:
LGTM

docs/command-line.asciidoc
tsg: We have them listed in https://www.elastic.co/guide/en/beats/packetbeat/current/_troubleshooting.html but they change too often to maintain the list, hence the weasel "list of commonly used debug selectors" wording in the doc :-)

ruflin: @tsg I see. Perhaps we could automate the generation of the list through searching for regexp pattern `logp.Debug("[a-zA-Z]"` in all files and then create a unique list. But not a priority.

monicasarbu:
@ruflin I think first we should clean them up a bit and keep only a few. I think they are too many. I would say to keep one selector per beat.

ruflin:
+1 for the cleanup

PR #166, opened 2015-10-19
logstash beats plugin updates

urso:
- Update lumberjack output to logstash beats plugin changes
- Update logstash test to use logstash-input-beats plugin


urso:
still dev preview until logstash-plugin-beats PR is merged. But seems to work in current state.

Issue #160, opened 2015-10-16
Update sample configuration files for all Beats


tsg:
They should show (commented out) the Logstash output and not the Redis one.

monicasarbu:
Instead of redis to put logstash as output.

andrewkroh:
Checkboxes for each beat that needs updated:
- [ ] Filebeat
- [ ] Packetbeat
- [ ] Topbeat

PR #167, opened 2015-10-19
Add logstash as output to the same config file of each beat

monicasarbu:


etc/libbeat.yml
```diff
@@ -98,27 +98,45 @@ output:
...
+
+    # Optional load balance the events between the Logstash hosts
+    # loadbalance: true
+
+    # Optional index name. The default index name is logstash
+    # index: logstash
+
+    # Optional TLS. The default is on
+    # tls:
+    #   disabled: true
```
andrewkroh: I think the other TLS options from ES should also be included here.

etc/libbeat.yml
```diff
@@ -121,6 +121,26 @@ output:
     #reconnect_interval: 1
 
 
+  ### Logstash as output
+
```
andrewkroh: Having this newline is inconsistent with the formatting of the others (elasticsearch, redis, file).

etc/libbeat.yml
```diff
@@ -121,6 +121,26 @@ output:
...
+    # Optional load balance the events between the Logstash hosts
+    #loadbalance: true
+
+    # Optional index name. The default index name is logstash
+    #index: logstash
+
+    # Optional TLS. The default is on
+    #tls:
+      #disabled: true
+
```
andrewkroh: I think the other TLS options from ES should also be included here.

etc/libbeat.yml
```diff
@@ -121,6 +121,26 @@ output:
...
+    # The Logstash hosts
+    #hosts: ["localhost:5044", "logstash:5045"]
+
+    # Optional load balance the events between the Logstash hosts
+    #loadbalance: true
+
+    # Optional index name. The default index name is logstash
+    #index: logstash
+
+    # Optional TLS. The default is on
```
tsg: I think I know what you are referring to, but the "the default is on" is missleading. If you leave everything disabled, tls is off, right?

andrewkroh:
Should redis be completely removed per #160?

etc/libbeat.yml
```diff
@@ -121,6 +121,26 @@ output:
...
+    #enabled: true
+
+    # The Logstash hosts
+    #hosts: ["localhost:5044", "logstash:5045"]
+
+    # Optional load balance the events between the Logstash hosts
+    #loadbalance: true
+
+    # Optional index name. The default index name is logstash
+    #index: logstash
```
tsg: The default is the beat name, not "logstash", see: https://www.elastic.co/guide/en/beats/libbeat/master/configuration.html#_index_2

etc/libbeat.yml
monicasarbu: the disabled: true option is tricky. It tricked me as well even if I knew it before.

monicasarbu:
Ready to merge.

PR #168, opened 2015-10-19
logstash integration test on based official logstash-input-beats plugin

urso:


scripts/docker/logstash/logstash.conf.2.tmpl
```diff
@@ -0,0 +1,16 @@
+input {
+  beats {
+    port => 12345
```
ruflin: should we use 5044 here too?

urso: let's do the port update in another PR. #169 already includes some changes. Missing ports will be updated after merging #168 and #169.

PR #187, opened 2015-10-21
Replace logstash with localhost

monicasarbu:


ruflin:
@monicasarbu I would suggest to just have 1 inside.

docs/configuration.asciidoc
```diff
@@ -175,7 +175,7 @@ output:
     enabled: true
 
     # The Elasticsearch cluster
-    hosts: ["http://10.45.3.2", "http://10.45.3.1/elasticsearch"]
+    hosts: ["http://localhost:9200"]
```
urso: why change these? Old entry did match comment better, not?

PR #190, opened 2015-10-21
Add table saying which Kibana versions are compatible with Elasticsearch

monicasarbu:


andrewkroh:
LGTM.

urso:
Isn't it up to kibana to tell which versions are supported?

monicasarbu:
@urso we want to make sure our users don't get an exception because of using the wrong Kibana version.

urso:
ok

PR #193, opened 2015-10-21
use min tls 1.0 if no version is configured

urso:


tsg:
@urso a test seems to be failing.

PR #195, opened 2015-10-21
Add beat->LS integration tests using TLS

urso:
we've found LS 1.5.4 still requires TLS 1.0, LS 2.x already supports up to TLS 1.2. For now integration tests run on LS 1.5.4 only, due to LS 1.5.4 based tests always installing most recent version of logstash-input-beats plugin.


.travis.yml
```diff
@@ -9,7 +9,7 @@ go:
 
 env:
   - TARGETS="check"
-  - TARGETS="testsuite ES_HOST=elasticsearch-173"
+  # - TARGETS="testsuite ES_HOST=elasticsearch-173"
```
urso: have had to disable elasticsearch 1.7 based integration tests. This PR doubles the number of LS->Elasticsearch tests. Thusly the already flaky tests became much more unstable.

ruflin: I created an issue here so we don't forget about enabling the tests again: https://github.com/elastic/libbeat/issues/196

PR #199, opened 2015-10-22
Mark Redis Output deprecated in the code

monicasarbu:
Add a warning message when using the redis output

### DIFF
--- a/outputs/redis/redis.go
+++ b/outputs/redis/redis.go
@@ -1,3 +1,6 @@
+// Starting with version 1.0.0-beta4 the Redis Output is deprecated as
+// it's replaced by the Logstash Output that has support for Redis Output plugin.
+
 package redis
 
 import (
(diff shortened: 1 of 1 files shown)


outputs/redis/redis.go
```diff
@@ -1,3 +1,6 @@
+// Starting with version 1.0.0-beta4 the Redis Output is deprecated as
```
ruflin: I suggest to add the word `@deprecated: ` in the beginning so it is very obvious. Otherwise it is hidden in the doc text.

outputs/redis/redis.go
```diff
@@ -66,6 +70,8 @@ type message struct {
 
 func (out *redisOutput) Init(beat string, config outputs.MothershipConfig, topology_expire int) error {
 
+	logp.Warn("Redis Output is deprecated. Please use the Redis Output Plugin from Logstash instead.")
```
ruflin: :+1:

PR #207, opened 2015-10-23
Publisher Preprocessor Logic Change

andrewkroh:
Change logic used by the publisher's preprocessor for determining whether to use `message.event` or `message.events`. This change allows the preprocessor to function correctly in the case that a client publishes an empty message.events. It also makes the logic consistent with other parts of the code like bulk.go.

I discovered that if I sent an empty events slice that the processor would complain about a `"Missing 'timestamp' field from event"` because it tried to send `message.event`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file based on the
 ### Backward Compatibility Breaks
 
 ### Bugfixes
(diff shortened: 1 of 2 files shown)


ruflin:
@andrewkroh Can you add it to the CHANGELOG.md?

andrewkroh:
I updated the changelog for this PR.

PR #216, opened 2015-10-26
elasticsearch output based on outputs/mode module

urso:
Remove retrying and connection pools logic from
output plugin.

Being based on outputs/mode number of clients/connections
to elasticsearch will not increase when elasticsearch
becomes unresponsive. Instead back pressure will be applied.

Resolves elastic/packetbeat#226


outputs/elasticsearch/api.go
```diff
@@ -310,67 +115,55 @@ func (es *Elasticsearch) Refresh(index string) (*QueryResult, error) {
...
 // Implements: http://www.elastic.co/guide/en/elasticsearch/reference/current/search-uri-request.html
-func (es *Elasticsearch) SearchURI(index string, docType string, params map[string]string) (*SearchResults, error) {
-
-	path, err := makePath(index, docType, "_search")
-	if err != nil {
-		return nil, err
-	}
-
-	resp, err := es.request("GET", path, params, nil)
+func SearchURI(es requestExecutor, index string, docType string, params map[string]string) (*SearchResults, error) {
```
ruflin: It looks like all this functions depend on requestExecutor. Would it make sense to created methods for requestExcutor?

outputs/elasticsearch/bulkapi.go
```diff
@@ -115,16 +56,19 @@ func (r *bulkRequest) Flush() (*QueryResult, error) {
...
+	index string, docType string,
+	params map[string]string, body []interface{},
+) (*QueryResult, error) {
+	return BulkWith(es, index, docType, params, nil, body)
 }
 
 // BulkWith creates a HTTP request containing a bunch of operations and send
 // them to Elasticsearch. The request is retransmitted up to max_retries before
 // returning an error.
-func (es *Elasticsearch) BulkWith(
```
ruflin: What was the driver to convert all the methods to functions?

outputs/elasticsearch/client.go
```diff
@@ -0,0 +1,295 @@
...
+)
+
+type Client struct {
+	elasticsearchConnection
+	index string
+}
+
+type elasticsearchConnection struct {
+	URL                string
+	Username, Password string
```
ruflin: Fore readability I would put this on two lines.

outputs/elasticsearch/client.go
```diff
@@ -0,0 +1,295 @@
...
+	"io"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/elastic/libbeat/common"
+	"github.com/elastic/libbeat/logp"
+)
+
+type Client struct {
```
ruflin: :+1: for the introduction of the client object.

outputs/elasticsearch/client.go
```diff
@@ -0,0 +1,295 @@
...
+		ts := time.Time(event["timestamp"].(common.Time))
+		index := fmt.Sprintf("%s-%d.%02d.%02d",
+			es.index, ts.Year(), ts.Month(), ts.Day())
+		meta := bulkMeta{
+			Index: bulkMetaIndex{
+				Index:   index,
+				DocType: event["type"].(string),
+			},
+		}
+		// meta := common.MapStr{
```
ruflin: Do you want to keep this comment in?

outputs/elasticsearch/client.go
```diff
@@ -0,0 +1,295 @@
...
+) (*QueryResult, error) {
+	return Index(es, index, docType, id, params, body)
+}
+
+func (es *Client) Refresh(index string) (*QueryResult, error) {
+	return Refresh(es, index)
+}
+
+func (es *Client) Delete(index string, docType string, id string, params map[string]string) (*QueryResult, error) {
+	return Delete(es, index, docType, id, params)
```
ruflin: Why not have the functionality for Delete directly in here instead of linking to a second function?

outputs/elasticsearch/client.go
```diff
@@ -0,0 +1,295 @@
...
+	r := &bulkRequest{
+		es:     es,
+		path:   path,
+		params: params,
+	}
+	r.enc = json.NewEncoder(&r.buf)
+	return r, nil
+}
+
+func (es *Client) sendBulkRequest(
```
ruflin: It is somehow confusing to read the code as `es` is used in different contexts (Client, elasticsearchConnection ,...)

outputs/elasticsearch/output.go
```diff
@@ -125,17 +142,81 @@ func (out *elasticsearchOutput) Init(
...
+		for _, host := range config.Hosts {
+			url, err := getURL(config.Protocol, config.Path, host)
+			if err != nil {
+				logp.Err("Invalid host param set: %s, Error: %v", host, err)
+			}
+
+			urls = append(urls, url)
+		}
+	} else {
+		// usage of host and port is deprecated as it is replaced by hosts
```
ruflin: I remember we decided to completely remove it for GA.

outputs/elasticsearch/output.go
```diff
@@ -230,116 +322,13 @@ func (out *elasticsearchOutput) UpdateLocalTopologyMap() {
...
-
-	// Assign default path if not set
-	if addr.Path == "" {
-		addr.Path = defaultPath
+func (out *elasticsearchOutput) randomClient() *Client {
+	switch len(out.clients) {
+	case 0:
+		return nil
+	case 1:
+		return out.clients[0].(*Client).Clone()
```
ruflin: What is the reason it is required to Clone the client here?

ruflin:
LGTM. Some comments added above to better understand why some parts of code were changed.

As this is quite a major change I would like to at least have on additional reviewer @tsg @monicasarbu @andrewkroh .

ruflin:
@urso Don't forget to update the CHANGELOG.md

outputs/elasticsearch/api.go
urso: requestExecutor has one method: request

outputs/elasticsearch/bulkapi.go
urso: Mostly driven by refactoring and having 'Elasticsearch' type and 'Client' side by side during refactoring + abstracting both via requestExecutor and bulkRequestSender. Made refactoring process more incremental having compilable version all the time.

In the end kinda liked having the interfaces (besides bad names), as they might allow changing strategies in future by implementing wrappers for requestExecute/bulkRequestSender (e.g. retrying multiple times) or testing (transport is decoupled from REST API call)

outputs/elasticsearch/client.go
urso: hehe, will do

outputs/elasticsearch/client.go
urso: ups, to be removed.

outputs/elasticsearch/client.go
urso: Yeah, copy'n paste. Will clean this up

outputs/elasticsearch/client.go
urso: mostly due to the introduction of the requestExecutor interface during refactoring. Places using these methods already have been easier to update by just wrapping the functions in methods.

outputs/elasticsearch/output.go
urso: ticket still open, that's why I left it in

outputs/elasticsearch/output.go
urso: possibly internal state on http or connection object when having an active request. I just don't expect client to be fully thread safe. connected field might be set by another thread and http module does quite some magic under the hood. Cloning copies the configuration only so we don't have to care about thread-safety at all.

urso:
From comments I'd say I have to do some more cleanup. Stay tuned.

outputs/elasticsearch/api.go
monicasarbu: I agree with @ruflin. It would be nice to group all these functions: Delete, SearchURI, etc as they are specific to an Elasticsearch Client.   Why not moving them to the Client object or to something new ?

outputs/elasticsearch/client.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// insert the events one by one
+	_, err := es.Index(index, event["type"].(string), "", nil, event)
+	if err != nil {
+		logp.Warn("Fail to insert a single event: %s", err)
+	}
+	return nil
+}
+
+func (es *Client) CreateIndex(index string, body interface{}) (*QueryResult, error) {
```
monicasarbu: Why not having the content of the entire CreateIndex (from api.go) in here. I find it a bit confusing to link it to a different function in the same dir.

outputs/elasticsearch/output.go
ruflin: :+1:

outputs/elasticsearch/output.go
ruflin: :+1:

outputs/elasticsearch/bulkapi.go
urso: functions are gone

outputs/elasticsearch/output.go
```diff
@@ -1,112 +1,103 @@
...
 	outputs.RegisterOutputPlugin("elasticsearch", elasticsearchOutputPlugin{})
 }
 
 type elasticsearchOutputPlugin struct{}
 
+type elasticsearchOutput struct {
+	index string
+	mode  mode.ConnectionMode
+
+	topology
```
ruflin: I don't like to have "topology" as part of each output as I think it should leave outside. But that is a discussion we can have later.

outputs/elasticsearch/output.go
```diff
@@ -125,130 +116,65 @@ func (out *elasticsearchOutput) Init(
...
-	// Ignore error here, as CreateIndex will error (400 Bad Request) if index
-	// already exists. If index could not be created, next api call to index will
-	// fail anyway.
-	_, _ = out.Conn.CreateIndex(".packetbeat-topology", nil)
-
-	setting := map[string]interface{}{
-		"server-ip": map[string]interface{}{
-			"_ttl": map[string]string{"enabled": "true", "default": "15s"},
-		},
+	out.TopologyExpire = 15000
```
ruflin: Ok with hardcoding this at the moment. Later we should use a constant.

ruflin:
LGTM

outputs/elasticsearch/output.go
urso: Just can not remove it yet

ruflin: :+1:

Issue #219, opened 2015-10-27
Change -e option to send the logging messages to stderr instead of stdout


monicasarbu:
Currently when the "-e" option is used in the command line of each beat, the logging messages are sent to stdout as it becomes easier to grep for them. If they would have been sent to stderr, then you need to use  -e 2>&1 | grep ... which in the end it's ok. 

Sending the logging messages to stderr seems to be a better approach here as it fulfills users' expectations.

Requested by: https://discuss.elastic.co/t/is-it-intended-to-log-errors-to-stdout-while-calling-the-option-in-the-code-stderr/33082

ruflin:
Closing based on https://github.com/elastic/libbeat/pull/221

PR #221, opened 2015-10-27
Use stderr for console log output

urso:


urso:
closes #219

logp/logp.go
```diff
@@ -107,7 +109,7 @@ func Init(name string, config *Logging) error {
 
 func SetStderr() {
```
andrewkroh: It looks like this function is not used anywhere? Should we just remove it.

andrewkroh:
LGTM

logp/logp.go
```diff
@@ -24,7 +24,7 @@ type Logging struct {
 func init() {
 	// Adds logging specific flags: -v, -e and -d.
 	verbose = flag.Bool("v", false, "Log at INFO level")
-	toStderr = flag.Bool("e", false, "Output to stdout and disable syslog/file output")
+	toStderr = flag.Bool("e", false, "Output to stderr and disable syslog/file output")
```
mrkschan: Shall this be "Log to stderr and disable syslog/file output"? Aligning to -v option.

urso: good catch

PR #223, opened 2015-10-27
configure beat log level

urso:


ruflin:
@tsg @urso With this, do we still need the -v flag?

urso:
I would keep -v in. It can still overwrite a restrictive log level from command line for some debugging

PR #225, opened 2015-10-27
make bulk publisher not send empty events

urso:


ruflin:
LGTM. Changelog?

urso:
It's a hidden improvement. Nothing was ever send, but reduces some annoying debug output. No real change -> didn't add a change log entry

PR #229, opened 2015-10-28
Add logging messages for bulk publishing in case of error

ruflin:
The goal here is to make it easier to spot problems with bulk publishing. This should users help to better understand what is going on.

@urso Have a look to see if this make sense from your perspective. I think the most important one to log is form BulkPublish as we just skipped the error so far.

### DIFF
--- a/outputs/mode/balance.go
+++ b/outputs/mode/balance.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/elastic/libbeat/common"
+	"github.com/elastic/libbeat/logp"
 	"github.com/elastic/libbeat/outputs"
 )
 
@@ -114,7 +115,7 @@ func (m *LoadBalancerMode) PublishEvent(
 
 func (m *LoadBalancerMode) publishEventsMessage(msg eventsMessage) error {
 	if ok := m.forwardEvent(m.work, msg); !ok {
-		outputs.SignalFailed(msg.signaler)
+		outputs.SignalFailed(msg.signaler, nil)
 	}
 	return nil
 }
@@ -167,7 +168,7 @@ func (m *LoadBalancerMode) start(clients []ProtocolClient) {
 func (m *LoadBalancerMode) onMessage(client ProtocolClient, msg eventsMessage) {
 	if msg.event != nil {
 		if err := client.PublishEvent(msg.event); err != nil {
-			m.onFail(msg)
+			m.onFail(msg, err)
 			return
 		}
 	} else {
@@ -184,7 +185,7 @@ func (m *LoadBalancerMode) onMessage(client ProtocolClient, msg eventsMessage) {
 				if send > 0 {
 					msg.attemptsLeft = m.maxAttempts + 1
 				}
-				m.onFail(msg)
+				m.onFail(msg, err)
 				return
 			}
 			published += n
@@ -194,9 +195,12 @@ func (m *LoadBalancerMode) onMessage(client ProtocolClient, msg eventsMessage) {
 	outputs.SignalCompleted(msg.signaler)
 }
 
-func (m *LoadBalancerMode) onFail(msg eventsMessage) {
+func (m *LoadBalancerMode) onFail(msg eventsMessage, err error) {
+
+	logp.Info("Error publishing events (retrying): %s", err)
+
 	if ok := m.forwardEvent(m.retries, msg); !ok {
-		outputs.SignalFailed(msg.signaler)
+		outputs.SignalFailed(msg.signaler, err)
 	}
 }
 
--- a/outputs/mode/failover.go
+++ b/outputs/mode/failover.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/elastic/libbeat/common"
+	"github.com/elastic/libbeat/logp"
 	"github.com/elastic/libbeat/outputs"
 )
 
@@ -109,8 +110,11 @@ func (f *FailOverConnectionMode) PublishEvents(
 ) error {
 	published := 0
 	fails := 0
+	var err error
+
 	for !f.closed && (f.maxAttempts == 0 || fails < f.maxAttempts) {
-		if err := f.connect(f.active); err != nil {
+		if err = f.connect(f.active); err != nil {
+			logp.Info("Connecting error publishing events (retrying): %s", err)
 			fails++
 			time.Sleep(f.waitRetry)
 			continue
@@ -121,6 +125,7 @@ func (f *FailOverConnectionMode) PublishEvents(
 			conn := f.conns[f.active]
 			n, err := conn.PublishEvents(events[published:])
 			if err != nil {
+				logp.Info("Error publishing events (retrying): %s", err)
 				break
 			}
 			published += n
@@ -140,7 +145,7 @@ func (f *FailOverConnectionMode) PublishEvents(
 		fails++
 	}
 
-	outputs.SignalFailed(trans)
+	outputs.SignalFailed(trans, err)
 	return nil
 }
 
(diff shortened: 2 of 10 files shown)


outputs/mode/balance.go
```diff
@@ -186,6 +187,8 @@ func (m *LoadBalancerMode) onMessage(client ProtocolClient, msg eventsMessage) {
 				}
 				m.onFail(msg)
 				return
+			} else {
+				logp.Info("Error publishing events: %s", err)
```
urso: here err == nil

urso: Maybe forward err to onFail which tries to forward the message to another worker. If forward fails (maxAttempts) SignalFailed is emitted. This is the place we finally drop the message.

urso: What about single events? See then-branch

outputs/mode/failover.go
```diff
@@ -121,6 +122,7 @@ func (f *FailOverConnectionMode) PublishEvents(
 			conn := f.conns[f.active]
 			n, err := conn.PublishEvents(events[published:])
 			if err != nil {
+				logp.Info("Error publishing events: %s", err)
```
urso: we retry multiple times. We want to log error for every failed attempt or only after last failed attempt?

urso: PublishEvent is used for single event. We want to add a log message too?

outputs/mode/single.go
```diff
@@ -71,6 +72,7 @@ func (s *SingleConnectionMode) PublishEvents(
 		for published < len(events) {
 			n, err := s.conn.PublishEvents(events[published:])
 			if err != nil {
+				logp.Info("Error publishing events: %s", err)
```
urso: same comments as in FailOver mode

outputs/mode/balance.go
ruflin: for err == nil, of course this needs to be changed.

I like the idea of forwarding it to onFail and having `onFail(msg eventsMessage, err error)`. Like this the logging could be done in one place.

We should do the same for single events. In case this pull request conflicts with https://github.com/elastic/libbeat/pull/216 we should merge yours first.

outputs/mode/failover.go
ruflin: As it is an "info" we should log it for every event. It would be good to have an "Err" message in case of full error. Where would have this to be added?

outputs/mode/single.go
ruflin: Lets do the discussion above :-)

outputs/mode/balance.go
urso: single events also use onFail ;)

outputs/mode/failover.go
urso: if you only want to log on once on failure it must be logged after outer loop. Search for outputs.SignalFailed for code locations sending definitely failed.

ruflin: I decided to log every time as we have the error here and we only log "info" not error.

outputs/mode/balance.go
```diff
@@ -194,7 +195,10 @@ func (m *LoadBalancerMode) onMessage(client ProtocolClient, msg eventsMessage) {
 	outputs.SignalCompleted(msg.signaler)
 }
 
-func (m *LoadBalancerMode) onFail(msg eventsMessage) {
+func (m *LoadBalancerMode) onFail(msg eventsMessage, err error) {
```
ruflin: @urso Switched to using onFail(...)

outputs/mode/balance.go
```diff
@@ -194,7 +195,10 @@ func (m *LoadBalancerMode) onMessage(client ProtocolClient, msg eventsMessage) {
...
+
+	logp.Info("Error publishing events: %s", err)
```
urso: maybe error message is misleading here, as we try to publish the event again. forwardEvent returns false if events are finally dropped.

urso: maybe have two error message. One for temp fail:

- Error publishing events (retrying): ...
- Error publishing events: ...

?

ruflin: So the message here could be: 

* Error publish events. Will retry)

And inside forwardEvent:

* Error publishing Event but this time with logp.Err(

outputs/mode/balance.go
```diff
@@ -194,8 +195,12 @@ func (m *LoadBalancerMode) onMessage(client ProtocolClient, msg eventsMessage) {
 	outputs.SignalCompleted(msg.signaler)
 }
 
-func (m *LoadBalancerMode) onFail(msg eventsMessage) {
+func (m *LoadBalancerMode) onFail(msg eventsMessage, err error) {
+
+	logp.Info("Error publishing events (retrying): %s", err)
+
 	if ok := m.forwardEvent(m.retries, msg); !ok {
+		logp.Err("Failing publishing events: %s", err)
```
urso: looks good. next failover and single modes

ruflin: ?

urso: Modes in failover.go and single.go are not finished yet, right?

ruflin: You mean we should also introduce onFail in single and failover?

urso: Not necessary onFail (put possible). But error reporting is not on par with balance.go

outputs/signal.go
```diff
@@ -140,7 +141,12 @@ func SignalCompleted(s Signaler) {
 }
 
 // SignalFailed sends the Failed event to s if s is not nil
-func SignalFailed(s Signaler) {
+func SignalFailed(s Signaler, err error) {
```
urso: we also got Signal function accepting an error parameter.

outputs/mode/failover.go
```diff
@@ -109,8 +110,10 @@ func (f *FailOverConnectionMode) PublishEvents(
 ) error {
 	published := 0
 	fails := 0
+	var err error = nil
```
urso: don't set to nil, it is already nil by default (linter configured in my editor complains)

outputs/mode/failover.go
urso: update log message?

    logp.Info("Error publishing events (retrying): %s", err)

same as in balance.go

outputs/mode/single.go
```diff
@@ -61,8 +62,12 @@ func (s *SingleConnectionMode) PublishEvents(
 ) error {
 	published := 0
 	fails := 0
+	var err error = nil
```
urso: remove = nil

outputs/mode/single.go
```diff
@@ -97,13 +103,20 @@ func (s *SingleConnectionMode) PublishEvent(
 	event common.MapStr,
 ) error {
 	fails := 0
+
+	var err error = nil
```
urso: no '= nil'

publisher/worker.go
```diff
@@ -87,6 +87,6 @@ func (ws *workerSignal) Init() {
 func stopQueue(qu chan message) {
 	close(qu)
 	for msg := range qu { // clear queue and send fail signal
-		outputs.SignalFailed(msg.signal)
+		outputs.SignalFailed(msg.signal, nil)
```
urso: optional:

    err := errors.New("Worker stopped")
    out.SignalFailed(msg.Signal, err)

maybe a bad idea, though

outputs/signal.go
```diff
@@ -149,10 +155,12 @@ func SignalFailed(s Signaler) {
 // Signal will send the Completed or Failed event to s depending
 // on err being set if s is not nil.
 func Signal(s Signaler, err error) {
+
 	if s != nil {
 		if err == nil {
 			s.Completed()
 		} else {
+			logp.Info("Failed to send event %s", err)
```
urso: maybe let's turn logic a little, as error will not be printed if s == nil.

    if err == nil {
        if s != nil {
            s.Completed()
        }
     } else {
         logp...
         if s != nil {
             s.Failed()
         }
    }

ruflin: I took a little bit a different route but output should be the same now.

PR #234, opened 2015-10-28
Configure number of workers in ES/LS outputer

urso:
configure number of workers per Host to be used by loadbalancing mode via
'worker' option

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -291,6 +291,11 @@ output:
 In the previous example, the Elasticsearch nodes are available at https://10.45.3.2:9220/elasticsearch and
 https://10.45.3.1:9230/elasticsearch.
 
+===== worker
+
+Number of workers per configured host publishing events to Elasticsearch. This
+is best used with load balancing mode enabled. Example: If you have 2 hosts and
+3 workers, in total 6 workers are started (3 for each host).
 
 ===== host (DEPRECATED)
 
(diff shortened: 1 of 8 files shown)


ruflin:
Do we have any recommendations for the number of workers? 1 worker per elasticsearch node? Needs rebasing again.

docs/configuration.asciidoc
```diff
@@ -291,6 +291,10 @@ output:
 In the previous example, the Elasticsearch nodes are available at https://10.45.3.2:9220/elasticsearch and
 https://10.45.3.1:9230/elasticsearch.
 
+===== worker
+
+Number of workers per configured host publishing events to Elasticsearch. Best
```
ruflin: I suggest to make this doc a little bit more extensive to better explain what a worker here exactly is.

urso: suggestions? It's just number of go-routines publishing to elasticsearch.

urso: See logstash doc: "The number of workers to use for this output" :)

ruflin: Worker defines the number of processes started to publish data to each host defined under hosts.

urso: Mmmhhh... this sounds like total number of workers for all hosts. Like if workers = 2 and I have 4 hosts only 2 processes are started. But it is per host, so 8 processes will be started.

ruflin: @urso That is a good explanation. It should be added as an example.

ruflin: @urso Following suggestion for the text: 

```
Number of workers per configured host publishing events to Elasticsearch. This is best used with load balancing mode enabled. Example: If you have 2 hosts and 3 workers, in total 6 workers are started (3 for each host).
```

urso: Like it.

docs/configuration.asciidoc
```diff
@@ -466,6 +471,12 @@ List of known logstash servers to connect to. All entries in this list can
 contain a port number. If no port number is given, the port options value is
 used as default port number.
 
+===== worker
+
+Number of workers per configured host publishing events to Elasticsearch. This
```
ruflin: Copy paste: Should be Logstash

PR #235, opened 2015-10-29
fail bulk request if an item index failed

### LINKED ISSUE #192: Item status code not checked by Elasticsearch Bulk API client
Summary: The Elasticsearch Bulk API client in libbeat does not check the status codes returned in the items array. This could result in data loss if an insert fails because the Beat will think the insert was property acknowledged and will not retry the delivery.

Background: The Elasticsearch Bulk API returns an "items" array containing one entry for every document that was indexed. There is a status code for each element. There is also an HTTP status code for the whole bulk API request. The HTTP response code can be 201 even though an item failed to insert.

Relevant Code: https://github.com/urso/libbeat/blob/master/outputs/elasticsearch/bulkapi.go#L42

Example response containing a failure:

```
{  
   "took":6477,
   "errors":true,
   "items":[  
      {  
         "create":{  
            "_index":"eventbeat-2014.07.24",
            "_type":"eventlog",
            "_id":"AVCMajFei1FaLysRhu6k",
            "_version":1,
            "status":201
         }
      },
      {  
         "create":{  
            "_index":"eventbeat-2014.07.24",
            "_type":"eventlog",
            "_id":"AVCMajFei1FaLysRhu6l",
            "status":429,
            "error":"EsRejectedExecutionException[rejected execution (queue capacity 50) on org.elasticsearch.action.support.replication.TransportShardReplicationOperationAction$PrimaryPhase$1@1420e113]"
         }
      }
   ]
}
```

urso:
Iterate bulk request response items and check status code.
On first status code >= 300 an error is returned with number
of successful send (and so far checked) items.

On failure the send modes will retry sending all events beginning
with first failed event. If bulk output mode is used, the remaining items
are forwarded to another worker for sending.

Closes #192

### DIFF
--- a/outputs/elasticsearch/client.go
+++ b/outputs/elasticsearch/client.go
@@ -70,6 +70,7 @@ func (client *Client) PublishEvents(
 		return events, ErrNotConnected
 	}
 
+	// new request to store all events into
 	request, err := client.startBulkRequest("", "", nil)
 	if err != nil {
 		logp.Err(
@@ -78,23 +79,11 @@ func (client *Client) PublishEvents(
 		return events, err
 	}
 
-	// encode events into bulk request buffer
-	var dropInit []int
-	for i, event := range events {
-		ts := time.Time(event["@timestamp"].(common.Time))
-		index := fmt.Sprintf("%s-%d.%02d.%02d",
-			client.index, ts.Year(), ts.Month(), ts.Day())
-		meta := bulkMeta{
-			Index: bulkMetaIndex{
-				Index:   index,
-				DocType: event["type"].(string),
-			},
-		}
-		err := request.Send(meta, event)
-		if err != nil {
-			logp.Err("Failed to encode event: %s", err)
-			dropInit = append(dropInit, i)
-		}
+	// encode events into bulk request buffer, dropping failed elements from
+	// events slice
+	events = bulkEncodePublishRequest(request, client.index, events)
+	if len(events) == 0 {
+		return nil, nil
 	}
 
 	// send bulk request
@@ -106,51 +95,76 @@ func (client *Client) PublishEvents(
 		return events, err
 	}
 
-	// check per item errors. On first failure encountered an error is returned
-	// with number of successful events published. Mode will not retry
-	// value reported ok until first failure was found.
-	var dropFail []int
-	softFails := 0
-	for i, rawItem := range res.Items {
-		status, err := itemStatus(rawItem)
-		if err == nil {
-			if status < 300 {
-				continue // ok value
-			}
-
-			if status < 500 && status != 429 {
-				// hard failure, drop element
-				dropFail = append(dropFail, i)
-				continue
-			}
-
-			softFails++
-		}
-	}
-
-	// report errors if some elements have been failed due to server errors
-	if softFails > 0 {
-		events = removedDropped(events, dropInit)
-		events = removedDropped(events, dropFail)
+	// check response for transient errors
+	events = bulkCollectPublishFails(res, events)
+	if len(events) > 0 {
 		return events, mode.ErrTempBulkFailure
 	}
 
 	return nil, nil
 }
 
-// drop elements given by indexes from events slice. Delete will not preserve order
-func removedDropped(events []common.MapStr, indexes []int) []common.MapStr {
-	if len(indexes) == 0 {
-		return events
-	}
+// fillBulkRequest encodes all bulk requests and returns slice of events
+// successfully added to bulk request.
+func bulkEncodePublishRequest(
+	requ *bulkRequest,
+	index string,
+	events []common.MapStr,
+) []common.MapStr {
+	okEvents := events[:0]
+	for _, event := range events {
+		ts := time.Time(event["@timestamp"].(common.Time))
+		index := fmt.Sprintf("%s-%d.%02d.%02d",
+			index, ts.Year(), ts.Month(), ts.Day())
+		meta := bulkMeta{
+			Index: bulkMetaIndex{
+				Index:   index,
+				DocType: event["type"].(string),
+			},
+		}
 
-	end := len(events) - 1
-	for _, i := range indexes {
-		events[i] = events[end]
-		end--
-	}
+		err := requ.Send(meta, event)
+		if err != nil {
+			logp.Err("Failed to encode event: %s", err)
+			continue
+		}
 
-	return events
+		okEvents = append(okEvents, event)
+	}
+	return okEvents
+}
+
+// bulkCollectPublishFails checks per item errors. On first failure encountered
+// an error is returned with number of successful events published. Mode will not
+// retry value reported ok until first failure was found.
+func bulkCollectPublishFails(
+	res *BulkResult,
+	events []common.MapStr,
+) []common.MapStr {
+	failed := events[:0]
+	for i, rawItem := range res.Items {
+		status, err := itemStatus(rawItem)
+		if err != nil {
+			logp.Info("Failed to parse bulk reponse for item (%i): %v", i, err)
+			// add index if response parse error as we can not determine success/fail
+			failed = append(failed, events[i])
+			continue
+		}
+
+		if status < 300 {
+			continue // ok value
+		}
+
+		if status < 500 && status != 429 {
+			// hard failure, don't collect
+			continue
+		}
+
+		debug("Failed to insert data(%i): %v", i, events[i])
+		logp.Info("Bulk item insert failed (%i)", i)
+		failed = append(failed, events[i])
+	}
+	return failed
 }
 
 func itemStatus(m json.RawMessage) (int, error) {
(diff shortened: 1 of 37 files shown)


urso:
Not sure if default behavior is ok. As @kimchy [explained](https://github.com/elastic/libbeat/issues/192#issuecomment-150053344) retrying with temporary sleep might be better.

One solution might be to mark the error as 'temporary' in special cases. If temporary error is returned, the worker (in load balancing mode) will back of retrying the (left over) bulk request after wait time (like fail-over mode already does).

andrewkroh:
With this change, after the first failure is encountered in the `items` array it is assumed that all the following elements are also failures. But failures can be interleaved with successes in the `items` arrays. So if the beat retries from the failure point it might actually be indexing duplicate events.

Here is a sample failure response (hopefully this helps): http://pastebin.com/raw.php?i=w2X79CkV

ruflin:
@andrewkroh I think returning on the first >= 300 is not a good idea exactly because of what you mentioned above. It can be that one entry in the middle of the bulk request is invalid but all others succeeded. Best would be to go through all events and return the count in the end on how many failed.

Resending all data can be somehow bad as it can duplicated hundreds of entries. At the same time hard to identify which is the exact entry which wasn't successful and only resend this one.

urso:
Yes, having failure in between is possible, that's why I wonder if it is ok to reindex the rest. Identifying exact entries and retrying logic is inter winded with output/mode module and needs some more adjustments. Still, if any entry in bulk fails too often, e.g. filebeat might retry to re-index complete bulk request, as granularity of error reporting to beat is not fine grained enough.

When introducing filtering we will have to push some options to output module to retry forever until all events have send (so events don't have to pass filter chain again). On the other hand we can introduce these kinds of options like right now (also improving filebeat shutdown) when sing becomes unresponsive.

andrewkroh:
As @urso mentioned in HipChat, moving some of the retry logic closer to the outputter will make it easier to handle retries.

This retry code should also take into account that there are two types of failures, those that are just intermittent (i.e. queue is full in ES) and those that are permanent (i.e data that is mismatched to the mapping). It won't help to retry the "hard" failures so we should just log an error and move on in that case.

urso:
to distinguis "hard" failures from "soft" failures, it's ok to use status code ranges?

    if < 300 {
        "success"
    } else if 400 <= status < 500 {
        "hard" fail
    } else {
        "soft" fail
    }

andrewkroh:
[429](http://httpstatusdogs.com/429-too-many-requests) is the exception to that rule.

References:
- https://github.com/elastic/elasticsearch/blob/148265bd164cd5a614cd020fb480d5974f523d81/core/src/main/java/org/elasticsearch/common/util/concurrent/EsRejectedExecutionException.java#L55
- https://www.elastic.co/blog/performance-considerations-elasticsearch-indexing

urso:
@andrewkroh oh right, thanks.

I think for now I would update the interface between mode and protocol client from return number of processed events to return a slice of to be processed events. Later one is more flexible and elasticsearch outputer can filter out events to be retried. Still doesn't change the fact retry policies should be given by publisher to output modes, as output modes can process them more efficiently.

andrewkroh:
I haven't reviewed the new code yet, but I did put it into action to test if this fixes the issue I was seeing with dropped events from eventbeat. 

Here is what happened while trying to publish about 24000 events from eventbeat. I had 6555 events successfully acknowledged from `PublishEvents(events, publisher.Sync)` and no errors returned. The _count from my Elasticsearch index shows 621755 events and it is still growing, so I'm guessing there is a large batch of events that are retrying resulting in duplicates.

I will review the code once I get some sleep.

urso:
Thanks for testing. I see I forgot to collect to remove already succeeded values in elasticsearch client response (only dropping hard failures).

andrewkroh:
Things are working better with these changes, but I'm still seeing more events in my index than what I published.

```
"publishedEvents": {"Application": 6984, "Internet Explorer": 0, "Security": 0, "System": 16907, "total": 23891}
```

```
GET /eventbeat-*/_count
{"count":61644,"_shards":{"total":775,"successful":775,"failed":0}}
```

publisher/output.go
```diff
@@ -30,22 +40,45 @@ func (o *outputWorker) onStop() {}
...
+		return
+	}
+
+	if o.maxBulkSize < 0 || len(events) <= o.maxBulkSize {
+		o.sendBulk(signal, events)
+		return
+	}
+
+	// start splitting bulk request
+	splits := (len(events) + (o.maxBulkSize - 1)) / o.maxBulkSize
```
andrewkroh: Small issue with the indexing used here. Here's how I changed it. Feel free to modify as you see fit.

```
	numSplits := (len(events) + (o.maxBulkSize - 1)) / o.maxBulkSize
	signal = outputs.NewSplitSignaler(signal, numSplits)
	for numEvents := len(events); numEvents > 0; numEvents = len(events) {
		splitIndex := o.maxBulkSize
		if numEvents < o.maxBulkSize {
			splitIndex = numEvents
		}

		o.sendBulk(signal, events[:splitIndex])
		events = events[splitIndex:]
	}
```

urso: thanks. should be fixed.

outputs/elasticsearch/client.go
```diff
@@ -64,44 +65,132 @@ func (client *Client) Clone() *Client {
...
+	events = bulkEncodePublishRequest(request, client.index, events)
+	if len(events) == 0 {
+		return nil, nil
+	}
+
+	// send bulk request
+	_, res, err := request.Flush()
+	if err != nil {
+		logp.Err(
+			"Failed to perform many index operations in a single API call: %s",
```
andrewkroh: This isn't new, but this message is not very precise IMO. What do you think of changing it to "Failed to perform _any_ bulk index operations."

outputs/elasticsearch/client.go
```diff
@@ -64,44 +65,132 @@ func (client *Client) Clone() *Client {
...
 	if !client.connected {
-		return 0, ErrNotConnected
+		return events, ErrNotConnected
 	}
 
+	// new request to store all events into
 	request, err := client.startBulkRequest("", "", nil)
 	if err != nil {
 		logp.Err(
 			"Failed to perform many index operations in a single API call: %s",
```
andrewkroh: This isn't new, but this message is not very precise IMO. What do you think of changing it to "Failed to perform _any_ bulk index operations."

outputs/elasticsearch/client.go
andrewkroh: The use of "many" seems pessimistic. Maybe it should say "Failed to perform some index operations in a bulk API call:" or "Some bulk index operations failed:".

outputs/elasticsearch/client.go
```diff
@@ -64,44 +65,132 @@ func (client *Client) Clone() *Client {
...
+			continue
 		}
+
+		okEvents = append(okEvents, event)
 	}
+	return okEvents
+}
+
+func eventBulkMeta(index string, event common.MapStr) bulkMeta {
+	ts := time.Time(event["@timestamp"].(common.Time))
```
andrewkroh: I think there should be a `.UTC()` conversion in there to guarantee the index date is based on UTC.

outputs/elasticsearch/client.go
```diff
@@ -64,44 +65,132 @@ func (client *Client) Clone() *Client {
...
+			Index:   index,
+			DocType: event["type"].(string),
+		},
+	}
+	return meta
+}
 
-	_, err = request.Flush()
+// bulkCollectPublishFails checks per item errors. On first failure encountered
+// an error is returned with number of successful events published. Mode will not
```
andrewkroh: This comment is out of date.

outputs/elasticsearch/output.go
```diff
@@ -19,16 +19,19 @@ var (
...
 const (
-	defaultEsOpenTimeout = 3000 * time.Millisecond
-
 	defaultMaxRetries = 3
 
-	elasticsearchDefaultTimeout = 30 * time.Second
+	elasticsearchDefaultTimeout = 90 * time.Second
 )
 
+var defaultBulkSize = 50
```
andrewkroh: Should this be declared as a `const` with the defaults above?

urso: I need the pointer to set in MothershipConfig. But having default const and allocate int where pointer is required might be cleaner solution

outputs/elasticsearch/client.go
urso: +1

publisher/bulk.go
```diff
@@ -108,8 +108,11 @@ func (b *bulkWorker) onEvents(signal outputs.Signaler, events []common.MapStr) {
 }
 
 func (b *bulkWorker) publish() {
+	// TODO: remember/merge and forward context options to output worker
```
andrewkroh: I think we should have a libbeat issue to track this enhancement.

andrewkroh:
This looks good to me. I just left a few minor cosmetic comments.

:white_circle: Note to self: retest load balance mode after merging and updating the beats.

tsg:
LGTM

ruflin:
Code LGTM. Didn't test if it runs as expected.

PR #236, opened 2015-10-29
Fix #227: Improve error message in run windows service

### LINKED ISSUE #227: Windows. The service process could not connect to the service controller.
I'm using filebeat as a log shipper and every time I am starting filebeat instance I'm receiving following error in console:

```
service_windows.go:45: ERR Error: The service process could not connect to the service controller.
```

I'm using 1.0.0-beta4 filebeat for windows.

My logstash.conf file:

```
input {
    beats {
        port => 5044
    }
}

output {
    stdout {}
}
```

My filebeat.yml file:

```
filebeat:
  prospectors:
    -
      paths:
        - "D:\\Repositories\\logstashtests\\logs.txt"
      type: log
output:
  logstash:
    enabled: true

    hosts: ["127.0.0.1:5044"]
```

For more information you can go to elastic/filebeat#145

ruflin:
Improve the error message to better explain what happens
Change the log level of the message to Info as it is not critical

### DIFF
--- a/outputs/logstash/logstash.go
+++ b/outputs/logstash/logstash.go
@@ -6,7 +6,6 @@ package logstash
 import (
 	"crypto/tls"
 	"errors"
-	"fmt"
 	"time"
 
 	"github.com/elastic/libbeat/common"
@@ -191,5 +190,4 @@ func (lj *logstash) addMeta(event common.MapStr) {
 		"beat": lj.index,
 		"type": event["type"].(string),
 	}
-	fmt.Printf("meta data: %v\n", event["@metadata"])
 }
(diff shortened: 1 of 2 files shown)


ruflin:
@tsg Can you have a look at this one? It doesn't directly solves the problem but should mitigate it. I'm not sure if there is an option to figure out if a service is run in the foreground or not.

tsg:
Looks like there can be a solution for this that avoids printing an error if we are in an "interactive" session.  I'll make another PR on the subject, so I'm closing this one. If I fail to make it work, we can still reopen this PR.

Issue #238, opened 2015-10-29
TLS insecure option is ignored for logstash


tsg:
It seems that the tls configuration is copied over in the transport [Connect](https://github.com/elastic/libbeat/blob/master/outputs/logstash/transport.go#L103) function. Not sure about the motivation behind this copy, but the problem is that the `InsecureSkipVerify` option is not copied over.

This breaks the insecure option. Should be an easy fix.

urso:
will add tests

tsg:
Test is tracked in #240.

PR #239, opened 2015-10-29
Fix insecure config option 

monicasarbu:
Trying to fix the issue https://github.com/elastic/libbeat/issues/238

### DIFF
--- a/outputs/logstash/transport.go
+++ b/outputs/logstash/transport.go
@@ -105,6 +105,7 @@ func (c *tlsClient) Connect(timeout time.Duration) error {
 	tlsconfig.RootCAs = c.tls.RootCAs
 	tlsconfig.Certificates = c.tls.Certificates
 	tlsconfig.ServerName = host
+	tlsconfig.InsecureSkipVerify = c.tls.InsecureSkipVerify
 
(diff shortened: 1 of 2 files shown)


ruflin:
LGTM

ruflin:
@monicasarbu Can you update the CHANGELOG?

outputs/logstash/transport.go
```diff
@@ -105,6 +105,7 @@ func (c *tlsClient) Connect(timeout time.Duration) error {
 	tlsconfig.RootCAs = c.tls.RootCAs
 	tlsconfig.Certificates = c.tls.Certificates
 	tlsconfig.ServerName = host
+	tlsconfig.InsecureSkipVerify = c.tls.InsecureSkipVerify
```
urso: For fixing mentioned bug this is enough, but other options might be missing (e.g. MaxVersion).

I think this is kind of a left over from before passing tls.Config type directly into client (has been forgotten during refactoring/cleanup). Replacing line 103 with:

    tlsconfig := c.tls

Should copy all configuration options.

alternatuce replace call line 113 with:

    socker := tls.Client(c.Conn, &c.tls)

Will replace in another PR including some tests

PR #249, opened 2015-11-02
single mode timeout config

urso:
- fix timeout configuration not applied to single mode
- add additional backoff mode to single mode on failure

Resolve elastic/filebeat#144


andrewkroh:
Testing notes: With `max_retries` set to `3` I am seeing CPU usage of 0.1%. With `max_retries` set to `1` I see about ~4%. If I disable `max_retries` by setting it to `-1` then I see high CPU usage because the spooler keeps retrying really fast.

andrewkroh:
LGTM

urso:
Test result is known limitation, as backoff is implemented in output plugin also retrying. If max_retries is disabled, the spooler will retry due to getting signaled an error almost immediately.

#235 makes progress to move 'sync' handling later in processing chain, but doesn't solve this one yet.

PR #252, opened 2015-11-03
Technical edits part2

dedemorton:
Mostly changes to make the content consistent and improve clarity.

@monicasarbu I'm still not sure whether I've gotten the description of name (for the shipper name) quite right with respect to how the network topology map works.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -1,8 +1,8 @@
 [[configuration]]
-== Configuration
+== Configuration Options
 
-The Beat configuration file uses
-http://yaml.org/[YAML] for its syntax. It consists of the following sections:
+The Beats configuration file uses
+http://yaml.org/[YAML] for its syntax. The file consists of the following sections:
 
 * <<configuration-shipper>>
 * <<configuration-output>>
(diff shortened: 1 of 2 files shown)


monicasarbu:
@dedemorton Yes, indeed. I am proposing to replace "shipper" with the "Beat" in the libbeat documentation when we speak about the Beat configuration and use the term "shipper" only in explaining what is a Beat. As it is a big change I would like to hear @tbragin @tsg thoughts about it. Thanks!

dedemorton:
@monicasarbu "Shipper" does appear quite a bit in the config file, though. Would you also update the config files, or just live with the disconnect?

ruflin:
@monicasarbu +1 for this one ;-)

monicasarbu:
@dedemorton We can easily update the comments in the configuration file, but changing the config options is not that easy as the config file needs to be backward compatible with the older version.

dedemorton:
@monicasarbu Yes of course. I was thinking about the comments. I do think it would be better if the terminology used in the doc is aligned with the terminology that is used in the comments in the config files, but it's probably not your highest priority. Just let me know how you want me to proceed with the doc changes.

monicasarbu:
@dedemorton I would say it's ok to replace shipper with the Beat. As it is now, it's hard to follow the libbeat documentation.

monicasarbu:
For replacing shipper with Beat in the docs, there is another issue #255 opened.

Issue #241, opened 2015-10-30
Update the latest stable version of Kibana and ES


monicasarbu:
Currently the documentation advise you to install Elasticsearch 1.7.2, Kibana 4.1.2 and Logstash 1.5.4. As Elasticsearch 2.0.0, Kibana 4.2.0 and Logstash 2.0.0 were released we need to update getting started guide (at least) with the newest versions.

* [x] update libbeat/index.asciidoc with the latest versions
* [x] check if the steps described in getting started guide still work (especially if the links are still working)

monicasarbu:
@dedemorton can you please have a look?

dedemorton:
@monicasarbu Quick question: is this change relevant to beta4, too? Just want to make sure that I am working in the correct branch. I'm working in the beta4 branch still with the assumption that everything is going to get pushed into Master. Is that correct.

monicasarbu:
@dedemorton yes, you can change the versions in beta4 as well.  All the docs related commits will be merged to Master today by @tsg.

dedemorton:
@monicasarbu  This is essentially resolved by PR #257. 

However, the deb and rpm steps need to be tested. And the paths for Elasticsearch will need to be updated  when the infra team resolves the related issue described here: https://github.com/elastic/infra/issues/548

So I will keep this open for now.

monicasarbu:
@dedemorton Thanks! We will do some testing for deb and rpm.

monicasarbu:
Close this issue for now. A new task (called Check getting started steps) is added to the things that need to be tested manually for each release, available in the `beats-qa` repository.

PR #257, opened 2015-11-04
implements changes for issue #241

dedemorton:
@monicasarbu I've updated the paths and tested the downloads. And I've moved the steps on starting Logstash to their own section because it doesn't make sense to run Logstash until after you've created the config file. I've used the same paths that we use on our download page.

I also updated the configuration file because the older settings were causing errors.

On Mac, I've tested all the installation steps (including the steps for all the Beats) to ensure that everything works.

I am hoping that someone else can test the installation steps on deb and rpm. I found a typo in one of the logstash commands, so they need testing.

One other thing: the very long path to the Elasticsearch download is causing the commands to wrap, so I've created an infra request to get a shorter path: https://github.com/elastic/infra/issues/548. When that issue is resolved, I'll update the paths to fix the display problem, but for now, at least the links work.

### DIFF
--- a/docs/gettingstarted.asciidoc
+++ b/docs/gettingstarted.asciidoc
@@ -34,7 +34,7 @@ deb:
 ["source","sh",subs="attributes,callouts"]
 ----------------------------------------------------------------------
 sudo apt-get install openjdk-7-jre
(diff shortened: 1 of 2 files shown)


monicasarbu:
@dedemorton Thank you for testing on OSX.

PR #261, opened 2015-11-04
fix setting ES default bulk size

urso:


monicasarbu:
Great that you found it.

PR #265, opened 2015-11-04
test tls insecure option for logstash output

### LINKED ISSUE #240: Create test for insecure option for logstash and elasticsearch
The goal is to test if the insecure option works as expected for elasticsearch and logstash output. See https://github.com/elastic/libbeat/issues/238 for more details.

urso:
resolves #240


ruflin:
LGTM. It would be nice to have also integration tests later for this.

PR #266, opened 2015-11-04
Respect max_retries=0

urso:
Output modes are configured with total number if send attempts, that is if
max_retries=0, total number of attempts = 1.

Also update output modes to handle failures and retries consistently for single
and bulk sends.

### DIFF
--- a/outputs/elasticsearch/output.go
+++ b/outputs/elasticsearch/output.go
@@ -90,6 +90,10 @@ func (out *elasticsearchOutput) init(
 	if config.Max_retries != nil {
 		maxRetries = *config.Max_retries
 	}
+	maxAttempts := maxRetries + 1 // maximum number of send attempts (-1 = infinite)
+	if maxRetries < 0 {
+		maxAttempts = 0
+	}
 
 	var waitRetry = time.Duration(1) * time.Second
 	var maxWaitRetry = time.Duration(60) * time.Second
@@ -98,15 +102,15 @@ func (out *elasticsearchOutput) init(
 	out.clients = clients
 	if len(clients) == 1 {
 		client := clients[0]
-		m, err = mode.NewSingleConnectionMode(client, maxRetries,
+		m, err = mode.NewSingleConnectionMode(client, maxAttempts,
 			waitRetry, timeout, maxWaitRetry)
 	} else {
 		loadBalance := config.LoadBalance == nil || *config.LoadBalance
 		if loadBalance {
-			m, err = mode.NewLoadBalancerMode(clients, maxRetries,
+			m, err = mode.NewLoadBalancerMode(clients, maxAttempts,
 				waitRetry, timeout, maxWaitRetry)
 		} else {
-			m, err = mode.NewFailOverConnectionMode(clients, maxRetries, waitRetry, timeout)
+			m, err = mode.NewFailOverConnectionMode(clients, maxAttempts, waitRetry, timeout)
 		}
 	}
 	if err != nil {
--- a/outputs/logstash/logstash.go
+++ b/outputs/logstash/logstash.go
@@ -99,18 +99,22 @@ func (lj *logstash) init(
 	if config.Max_retries != nil {
 		sendRetries = *config.Max_retries
 	}
+	maxAttempts := sendRetries + 1
+	if sendRetries < 0 {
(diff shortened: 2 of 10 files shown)


tsg:
@urso Please also update docs and changelog.

tsg:
Code looks good to me.

outputs/mode/single.go
```diff
@@ -65,47 +65,19 @@ func (s *SingleConnectionMode) PublishEvents(
...
-		}
-
-	sendFail:
-		logp.Info("send fail")
-		backoff := time.Duration(int64(s.waitRetry) * (1 << backoffCount))
-		if backoff > s.maxWaitRetry {
-			backoff = s.maxWaitRetry
-		} else {
-			backoffCount++
+			// XXX: what if subset of events have been send? reset failure count
```
andrewkroh: To answer the comment, I think resetting the failure count in this case would be consistent with the behavior defined in other parts of the code. Right?

outputs/mode/failover.go
```diff
@@ -153,24 +128,53 @@ func (f *FailOverConnectionMode) PublishEvent(
...
+			logp.Info("Error publishing events (retrying): %s", err)
+			return false, false
+		}
+		return true, false
+	})
+}
+
+func (f *FailOverConnectionMode) publish(
+	signaler outputs.Signaler,
+	fn func() (bool, bool),
```
andrewkroh: For maintainability, I think it would be great to document what `fn`'s return values mean. Here's my attempt:

```
// publish sends the events to the outputter and signals success or failure to signaler.
// fn is called to send the events to the outputter and it returns (bool, bool). The first
// bool indicates if the send was OK and second indicates if the failure count should be
// reset. Failure count is typically reset when there is a partial send.
```

andrewkroh:
LGTM

outputs/elasticsearch/output.go
```diff
@@ -90,6 +90,7 @@ func (out *elasticsearchOutput) init(
 	if config.Max_retries != nil {
 		maxRetries = *config.Max_retries
 	}
+	maxAttempts := maxRetries + 1 // maximum number of send attempts (-1 = infinite)
```
urso: set maxAttempts = 0 if maxRetries < 0 . if e.g. maxRetries is configured to -2 modes might misbehave (assuming values >= 0)

outputs/logstash/logstash.go
```diff
@@ -99,18 +99,19 @@ func (lj *logstash) init(
 	if config.Max_retries != nil {
 		sendRetries = *config.Max_retries
 	}
+	maxAttempts := sendRetries + 1
```
urso: see above comment if sendRetries < 0

outputs/mode/single.go
urso: right

outputs/elasticsearch/output.go
urso: fixed

outputs/logstash/logstash.go
urso: fixed

outputs/mode/failover.go
```diff
@@ -153,24 +132,61 @@ func (f *FailOverConnectionMode) PublishEvent(
...
+// It provides general error handling and back off support used on failed
+// send attempts. To be used by PublishEvent and PublishEvents.
+// The send callback will try to progress sending traffic and returns kind of
+// progress made in ok or resetFail. If ok is set to true, send finished
+// processing events. If ok is false but resetFail is set, send was partially
+// successful. If send was partially successful, the fail counter is reset thus up
+// to maxAttempts send attempts without any progress might be executed.
+func (f *FailOverConnectionMode) publish(
+	signaler outputs.Signaler,
+	fn func() (ok bool, resetFail bool),
```
andrewkroh: I think you meant to rename `s/fn/send/` like you did to the other `fn`.

andrewkroh:
Looks good to me (though these type of changes are hard to look at and say they are correct).

urso:
removed review. I'd like to add some tests for error handling, which is quite delicate.

tsg:
Thanks @urso for adding tests, makes me feel more confident in taking this into rc1 :-).

PR #278, opened 2015-11-06
switch to glide

urso:
experimental (WIP) PR for switching from godep to glide vendoring tools.


tsg:
Will be done in the merged repo.

Issue #280, opened 2015-11-09
Beats Standardisation and Update Automation


ruflin:
Every beat has a set of files and dependencies in common. The goal of this document is to discuss two issues:

* What is the easiest way to start a new beat? Which files are needed?
* How can an existing beat kept up to date? How can we make it easy that a beat profits from the newest enhancements to libbeat and the tooling around?

# Create a New Beat
A new beat can be generated in two ways. Either an existing simple beat like topbeat is copied or a generator is used to created the basic required files.

The following is a list which all beats should have in common. At least part of these files should be identical across all beats.

* .editorconfig
* .gitattributes
* .gitignore
* .travis.yml
* beat/yourbeat.go
* CHANGELOG.md
* config/config.go
* docs/index.asciidoc
* etc/Xbeat.yml
* etc/fields.yml
* Godeps/.../libbeat/
* LICENSE
* main.go
* Makefile
* README.md
* pre commit hook

In addition there are scripts / libraries from libbeat which currently exist in all beats but could be reused:

* scripts/crosscompile.sh
* scripts/generate-*.py

A generator could be used to create the above files and fill them with the standard content based on the beat name.

# Updating Existing Beat
A bigger challenge is to update existing beats. As shown above, lots of files are shared and parts of these files are identical. A fix in the Makefile or an improvement in the .gitignore file can be beneficial for all beats, but at the same time every beat has its own additions. In case the libbeat base config is updated, all the depending beats must be updated. Updating Godeps only updates the go files.

To simplify this process and make it scalable, it should be automated.

# Proposed Solution

For the problem "create beat" I think the best solution is to clone an existing simple beat. But to do this is must be ensured that the beat is up-to-date with the most recent changes and can fetch new changes. This directly requires a solution for problem number two.

## Template Line

A differentiation must be made between the files which must be updated and the ones which only must be generated. README.md should never be overwritten, but .gitignore parts should be added if new ones exist in libbeat. All files which must be updated, should be split into two parts: beat specifics, libbeat details. To make it possible to use the "original" files but still make it modifyable, a standard "template" line should be introduced. For example the .gitignore file could look as following:

```
file.test
tmp

### libbeat generated - do not change ###

build
.vagrant
...
```

Then a script would be used to automatically replace everything after the headline with the most recent changes from libbeat if called.

The same would apply to all other files which would have to updated.

## Automatic Branch and Committing
[...]

PR #284, opened 2015-11-10
Standardise Config and Files across Beats

ruflin:
This make automatic replacement possible.

Remove duplicated save_topology

See also https://github.com/elastic/filebeat/pull/218


ruflin:
Related to https://github.com/elastic/libbeat/issues/280

Makefile
```diff
@@ -142,11 +144,16 @@ coverage-report:
 	tail -q -n +2 ./${COVERAGE_DIR}/*.cov >> ./${COVERAGE_DIR}/full.cov
 	$(GODEP) go tool cover -html=./${COVERAGE_DIR}/full.cov -o ${COVERAGE_DIR}/full.html
 
+# Update expects the most recent version of libbeat in the GOPATH
+.PHONY: update
+update:
+	bash ${GOPATH}/src/github.com/elastic/libbeat/scripts/update.sh ${BEATNAME} github.com/elastic/${BEATNAME}
+
```
tsg: This is meant to be executed in the Beat, rather than in libbeat, right? Because I executed `make update` and it seemed to mostly delete some files.

scripts/update.sh
```diff
@@ -0,0 +1,59 @@
...
+cat ${LIBBEAT}/.editorconfig > .editorconfig
+
+echo "Update .gitattributes"
+cat ${LIBBEAT}/.gitattributes > .gitattributes
+
+echo "Update LICENSE"
+cat ${LIBBEAT}/LICENSE > LICENSE
+
+echo "Update MAKEFILE"
+cat ${LIBBEAT}/Makefile > scripts/Makefile
```
tsg: Shouldn't we use another "template" Makefile instead of the root one from libbeat? I imagine we'll have libbeat specific things in this one.

scripts/update.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
```
tsg: I recommend having `set -e` in this script so we catch errors in case there are any.

tsg:
Generally LGTM, but I tried to run `make update` in filebeat (both filebeat and libbeat in your branches), and it just removed the files.

Makefile
ruflin: Yes, it is meant to be executed in a beat. Of course it should not delete files :-(

scripts/update.sh
ruflin: Added

scripts/update.sh
ruflin: My initial idea was, that everything we do for libbeat, should be the minimum for other beats. That is why I used the main file. So far it works quite well but probably having the same structure as in all the other beats is easier to understand and will make it more extendable. I will change it.

andrewkroh:
How about a standardized [pre-commit](https://github.com/elastic/packetbeat/blob/master/pre-commit) hook for each beat? I think only packetbeat provides one now. Or maybe we should just remove packetbeat's and not have one.

andrewkroh:
I look forward to not having to hand update each beat's config file. :+1: 

This isn't necessarily a blocker, but I wanted to point out that if someone has multiple paths in their `GOPATH` environment variable (i.e. `GOPATH=/this/go:/that/go`) then any place that directly uses $GOPATH to create a path will fail. One workaround might be to use something like `echo $GOPATH | tr ':' '\n' | head -1` to get only the first path.

ruflin:
@andrewkroh I like the idea of the pre-commit hook. I added it here to the list. https://github.com/elastic/libbeat/issues/280

ruflin:
About GOPATH: I was hoping that most people only have 1 (outside Godep). But you are right this will break the script and also our build script. Here we have the same issue:https://github.com/elastic/libbeat/blob/master/Makefile#L5

tsg:
Just in case I won't have time to review further here, feel free to merge this one even if I don't respond to comments.

ruflin:
More refactoring done and inputs added. @andrewkroh @urso I think it is now at a good stage for additional review / merging.

publisher/worker_test.go
```diff
@@ -8,6 +8,8 @@ import (
 
 // Test sending events through the messageWorker.
 func TestMessageWorkerSend(t *testing.T) {
+	t.Skip("Skip because or race condition problem")
```
andrewkroh: If you add `// +build !race` to the top of the file it will disable these test from running with the race detector, but the test will still run if someone runs just the `unit-tests`.

ruflin: I hope actually that this skipping is only temporary. Can the !race be overwritten by -race?

scripts/Makefile
```diff
@@ -0,0 +1,180 @@
...
+	go get github.com/tools/godep
+	mkdir -p ${BUILD_DIR}/bin
+	source $(GOPATH)/src/github.com/elastic/libbeat/scripts/crosscompile.bash; OUT='${BUILD_DIR}/bin' go-build-all
+
+# Fetch dependencies
+.PHONY: deps
+deps:
+	go get github.com/tools/godep
+	# TODO: Is this still needed?
+	go get -t ./...
```
andrewkroh: I think this should be removed since all dependencies should be coming from `Godeps`.

scripts/Makefile
```diff
@@ -0,0 +1,180 @@
...
+# Create test coverage binary
+.PHONY: libbeat.test
+libbeat.test: $(GOFILES)
+	$(GODEP) go test -c -covermode=atomic -coverpkg ./...
+
+# Cross-compile beat for the OS and architectures listed in
+# crosscompile.bash. The binaries are placed in the ./bin dir.
+.PHONY: crosscompile
+crosscompile: $(GOFILES)
+	go get github.com/tools/godep
```
andrewkroh: This target can just depend on `deps` like the `build` target does instead of directly calling `go get`.

However, my opinion is to not have `build` or `crosscompile` explicitly depend on `deps`, I would prefer have them independent because of speed, and also because I occasionally build while offline. 

Sometimes `go get` is really slow...
```
$ time go get github.com/tools/godep

real	0m12.997s
user	0m0.200s
sys	0m0.300s
```

ruflin: As we use `go get github.com/tools/godep` no internet connection is required if the package already exists locally. So the above should also work without internet connection and should be very fast if it is there:

```
 time go get github.com/tools/godep

real    0m0.137s
user    0m0.110s
sys     0m0.045s
```

scripts/Makefile
ruflin: I removed it now.

publisher/worker_test.go
ruflin: Adjusted it accordingly to your recommendation.

scripts/Makefile
```diff
@@ -0,0 +1,178 @@
...
+# Create test coverage binary
+.PHONY: libbeat.test
+libbeat.test: $(GOFILES)
+	$(GODEP) go test -c -covermode=atomic -coverpkg ./...
+
+# Cross-compile beat for the OS and architectures listed in
+# crosscompile.bash. The binaries are placed in the ./bin dir.
+.PHONY: crosscompile
+crosscompile: deps $(GOFILES)
+	go get github.com/tools/godep
```
andrewkroh: Now it checks for godep twice. ;)

ruflin: To speed it up ;-) Fixed

scripts/update.sh
```diff
@@ -0,0 +1,61 @@
...
+echo "Update .gitattributes"
+cat ${LIBBEAT}/.gitattributes > .gitattributes
+
+echo "Update LICENSE"
+cat ${LIBBEAT}/LICENSE > LICENSE
+
+echo "Update MAKEFILE"
+cat ${LIBBEAT}/scripts/Makefile > scripts/Makefile
+sed -i "" -e s/libbeat.test/$BEATNAME.test/g scripts/Makefile
```
mrkschan: Why do we choose `libbeat` as the placeholder but not using an explicit `placeholder`? 

E.g. `sed -i "" -e s/PLACEHOLDER.test/$BEATNAME.test/g scripts/Makefile`

ruflin: The main reason is because there is no replacement inside the libbeat repo. It is the Makefile that is used by libbeat directly. If we use placeholders, we have to add one more file to the libbeat repo and would have to run the generation script also inside libbeat.

PR #285, opened 2015-11-10
Cleanup the structure for the fields that identify a beat

### LINKED ISSUE #281: Consolidate beat fields (hostname, name, ip, etc.)
### Current state

First, to summarize the current situation:

* Currently libbeat adds the `shipper` field in all beats, which by default is equal to the host name on which the beat is running on, but it can be overwritten with anything the user wants from the configuration file.

* Logstash-Forwarder used to include the hostname in a `host` field. There is this Filebeat [ticket](https://github.com/elastic/filebeat/issues/208) about losing the compatibility with LSF from this point of view.

This issue has made us realize that the `shipper` field is confusing and the behavior of defaulting to the hostname but make it possible to change it makes it difficult to use across the different beats. In addition, we realized that in the future we are likely to need more fields to identify a beat, like a UUID or the IP address for where it's sending data.

### Plan for the future

Our ideal future situation is that each Beats exports several related fields grouped in a "beat" sub-dictionary:

* beat
  * name (by default the hostname, but changable from the configuration file)
  * hostname (old default shipper value) -> cannot be overwritten in config
  * version
  * uuid
  * ip 
  * type (packetbeat/filebeat/topbeat)

The sub-dictionary is there to group the fields together but also to avoid conflicts with the fields of various Beats. For example, Packetbeat already has and `ip` field.

### Plan for the next release

For `1.0.0-rc2`/`1.0.0-ga`, we plan the following changes:

 * Add the `beat.hostname` field, always set to the hostname as returned by the operating system
 * Rename `shipper` into `beat.name`l.
 * Add the `beat.version` field, because it's easy to add and we are sure it will be useful.

The rest of the fields we don't want to add for GA because we're not yet sure about the format or if they are actually needed. The IP was excluded for now because of difficulties in selecting the right IP address in case more are configured. But since we have the "beat" namespace reserved, we can easily add them later.

The `beat.hostname` is not directly compatible with the LSF, but it can be easily renamed via a Logstash filter when required.

tsg:
* The `shipper` field is renamed to `beat.name`
* A `beat.hostname` field is added that contains the hostname
* A `beat.version` field is added that contains the version

Closes #281.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,15 @@ All notable changes to this project will be documented in this file based on the
 
 ### Backward Compatibility Breaks
 
+* The `shipper` output field is renamed to `beat.name`. #285
+
 ### Bugfixes
 
-
 ### Added
+
+* Added `beat.hostname` to contain the hostname where the Beat is running on as
+  returned by the operating system. #285
+* Added `beat.version` to contain the version of the Beat. #285
 
 ### Deprecated
 
--- a/beat/beat.go
+++ b/beat/beat.go
@@ -93,7 +93,7 @@ func (b *Beat) LoadConfig() {
 
 	logp.Debug("beat", "Initializing output plugins")
 
-	if err := publisher.Publisher.Init(b.Name, b.Config.Output, b.Config.Shipper); err != nil {
+	if err := publisher.Publisher.Init(b.Name, b.Version, b.Config.Output, b.Config.Shipper); err != nil {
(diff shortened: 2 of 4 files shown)


tsg:
This will require changes in the Beats as well (at least etc/fields.yml additions), but I'll do those after this one gets merged for a clean process.

ruflin:
We should probably later also update the docs / migration guide.

PR #286, opened 2015-11-11
Fix logging issues.

### LINKED ISSUE #275: Debian logging to file fails if path not configured
Initialization of file logger fails (tested on debian 8) if no path is configured.

On linux default path is /var/log/<beatname>, but startup fails if logging.files.path is not configured:

    # filebeat -c /etc/filebeat/filebeat.yml
    Error initializing logging: mkdir : no such file or directory

Setting path to default value (/var/log/<beatname> ) in config file resolves problem and directory is generated.

### LINKED ISSUE #276: Logs are always written to stderr
Even if logging to stderr is not explicitly configured on command line, logs are still written to stderr.

On startup (before setting up) stderr is enabled by default: https://github.com/elastic/libbeat/blob/master/logp/logp.go#L88

SetStderr must be called to disable stderr logging after setup. This is only done by packetbeat, but no other beat.

andrewkroh:
Fixes #275.
Fixes #276.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ All notable changes to this project will be documented in this file based on the
 * The `shipper` output field is renamed to `beat.name`. #285
 
 ### Bugfixes
+- Disable logging to stderr after configuration phase. #276
+- Set the default file logging path when not set in config. #275
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


tsg:
LGTM.

Issue #264, opened 2015-11-04
Proposal: Simply configuration options


andrewkroh:
I noticed a few areas of the configuration files that I think could be simplified. This is a proposal to make some changes and I would like to get feedback prior to changing anything.

1. Remove boolean `enabled` and `disabled` options. Use commenting as an alternative.

2. Explicitly declare a logging output with `logger: syslog` or `logger: file` or `logger: [syslog, file]`. Currently loggers are configured as multiple boolean options (i.e. `to_syslog: true`, `to_files: true`).

3. Rename the command line option of `-test` to `-configtest`. The more clearly identifies the purpose of the flag and aligns with logstash's analogous flag.

urso:
+1 for points 1 and 3

point 2 won't work well with yaml unless we start parsing ourselves. Would rather like to see:

    logging:
        syslog:

        file:
            ...

andrewkroh:
I agree, it would be better to do it as @urso suggested. It is also more consistent with how the outputs are configured.

I added a forth proposed change above.

mrkschan:
Can we have (1) asap?

ruflin:
I'm +1 on 1-4, but for the 2 the suggestion from @urso as it will allow more logging specific configs (if needed).

PR #295, opened 2015-11-12
Removing enabled as a configuration option for outputs.

andrewkroh:
Removing disabled as a configuration opation for tls configuration.
Use comments in config file to enable and disable. #264

Suggested diff view: [Ignore trailing whitespace](https://github.com/elastic/libbeat/pull/295/files?w=1) (the docs had a lot of trailing whitespace)


urso:
LGTM

ruflin:
LGTM.

It's good that you marked it as BC break as one issue I see popping up, is that some users will have uncommented the output lines and as enabled / disabled is not relevant anymore, it will active the output. We should elaborate on that in the release blog post.

PR #296, opened 2015-11-13
Bug/226 logstash record loss

urso:
Resolves elastic/filebeat#226

- make logstash window size configurable based on bulk_max_size config file option
- fix integration test testSendMultipleBigBatchesViaLogstash which was supposed to find this kind of issue (after fix error was reproducible)
- refactor logstash publishing to correctly return unprocessed events on failures only

### DIFF
--- a/outputs/logstash/client.go
+++ b/outputs/logstash/client.go
@@ -73,8 +73,22 @@ func (l *lumberjackClient) PublishEvent(event common.MapStr) error {
 func (l *lumberjackClient) PublishEvents(
 	events []common.MapStr,
 ) ([]common.MapStr, error) {
+	for len(events) > 0 {
+		n, err := l.publishWindowed(events)
+		events = events[n:]
+		if err != nil {
+			return events, err
+		}
+	}
+	return nil, nil
+}
+
+// publishWindowed published events with current maximum window size to logstash
+// returning the total number of events send (due to window size, or acks until
+// failure).
+func (l *lumberjackClient) publishWindowed(events []common.MapStr) (int, error) {
 	if len(events) == 0 {
-		return nil, nil
+		return 0, nil
 	}
 
 	// prepare message payload
@@ -83,7 +97,7 @@ func (l *lumberjackClient) PublishEvents(
 	}
 	count, payload, err := l.compressEvents(events)
 	if err != nil {
-		return events, err
+		return 0, err
 	}
 
 	if count == 0 {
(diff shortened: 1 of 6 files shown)


outputs/logstash/client.go
```diff
@@ -31,7 +32,6 @@ type lumberjackClient struct {
 // TODO: make max window size configurable
```
andrewkroh: Is that comment now wrong?

urso: yep.

andrewkroh:
LGTM. Nice job on tracking down the issue so quickly.

ph:
Glad its fixed!

outputs/logstash/client.go
```diff
@@ -68,8 +72,22 @@ func (l *lumberjackClient) PublishEvent(event common.MapStr) error {
 func (l *lumberjackClient) PublishEvents(
```
ruflin: Can you add a doc block here that this will return the events which were not sent?

outputs/logstash/logstash_integration_test.go
```diff
@@ -113,11 +113,14 @@ func testElasticsearchIndex(test string) string {
 }
 
 func newTestLogstashOutput(t *testing.T, test string, tls bool) *testOutputer {
+	windowSize := 32
```
ruflin: Even though we set here the windowSize to 32, I suggest to add a second integration test with leaving the default window size (what most people have) and have a log file without about 10k entries and sending it. Just to be sure :-)

ruflin:
LGTM. See comments and needs rebase.

ruflin:
@urso I opened a pull request to your branch with the suggested changes: https://github.com/urso/libbeat/pull/1

outputs/logstash/client.go
urso: here is the full description of the interface:

https://github.com/elastic/libbeat/blob/master/outputs/mode/mode.go#L51

outputs/logstash/logstash_integration_test.go
urso: I don't think that's necessary. In your PR you created a test with batch size > window size. But the current test is already testing batch sizes > window size.

Plus, bug was also triggered by batch size < window size. Will add another test with batch size < window size.

outputs/logstash/client.go
urso: done

PR #299, opened 2015-11-13
Added timestamp for file logging. 

### LINKED ISSUE #291: Timestamp and caller are missing when logging to file
Log output to stderr looks as following:

```
2015/11/12 09:03:37.369262 geolite.go:52: WARN Couldn't load GeoIP database
```

File output for the same event:
```
WARN Couldn't load GeoIP database
```

For stderr, the timestamp is set through flags (https://github.com/elastic/libbeat/blob/master/logp/log.go#L142) so it is automatically generated afterwards. Same for syslog. For file output it must be generated manually. The applies to the caller reference.

For reference also see: https://github.com/elastic/filebeat/issues/223

ruflin:
Closes  #291

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -848,6 +848,15 @@ The number of most recent rotated log files to keep on disc. Older files are
 deleted during log rotation. The default value is 7. The keepfiles options has to be
 in the range of 2 to 1024 files.
 
+==== Logging Format
+
+The logging format is different for each logging type. In case of syslog, syslog adds the timestamp itself.
+For logging to file, the format RFC3339 is used which is `2006-01-02T15:04:05Z07:00 WARN log-message`.
+First is the timestamp with the time zone, then
+the logging level. As logging to stderr should be mainly used for debugging, additional debug information
+is added here. The format is `2015/11/12 09:03:37.369262 geolite.go:52: WARN log-message`. First is the UTC timestamp
+with the milliseconds, then the name of the caller that sent the log entry followed by the log level.
+
 [[configuration-run-options]]
 === Run options (optional)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file based on the
 * Use of `enabled` as a configuration option for outputs (elasticsearch,
   logstash, etc.) has been removed. #264
 * Use of `disabled` as a configuration option for tls has been removed. #264
+- The `shipper` output field is renamed to `beat.name`. #285
 
 ### Bugfixes
 - Disable logging to stderr after configuration phase. #276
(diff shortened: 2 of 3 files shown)


docs/configuration.asciidoc
```diff
@@ -848,6 +848,14 @@ The number of most recent rotated log files to keep on disc. Older files are
 deleted during log rotation. The default value is 7. The keepfiles options has to be
 in the range of 2 to 1024 files.
 
+==== Logging Format
+
+The logging format is different for each logging type. In case of syslog, syslog adds the timestamp itself.
+For logging to file, the format `2015-11-13 08:52:29 WARN log-message` is used. First ist the UTC timestamp, then
```
urso: s/ist/is/

andrewkroh:
You changes look good. I noticed two other things:

- I think syslog should also receive the filename:lineno prefix.
- I think all messages should contain their level (INFO, WARN, DEBUG). Currently only debug messages get DBG.

I can make these changes in a separate PR if you all agree.

docs/configuration.asciidoc
```diff
@@ -848,6 +848,15 @@ The number of most recent rotated log files to keep on disc. Older files are
...
 in the range of 2 to 1024 files.
 
+==== Logging Format
+
+The logging format is different for each logging type. In case of syslog, syslog adds the timestamp itself.
+For logging to file, the format RFC3339 is used which is `2006-01-02T15:04:05Z07:00 WARN log-message`.
+First is the timestamp with the time zone, then
+the logging level. As logging to stdErr should be mainly used for debugging, additional debug information
+is added here. The format is `2015/11/12 09:03:37.369262 geolite.go:52: WARN log-message`. First is the UTC timestamp
+with the milliseconds, then the name of the caller that sent the log entry followed by the log level.
```
urso: what happened here? timezone in second example wrong.

ruflin: StdErr output was always in UTC. Should we change that?

ruflin:
@andrewkroh For me the "caller" info is more for debugging, that's why I like it in  stdErr but not file or syslog. For the second one on the warn level: Yes, all should defintively include it. You can also push it into this PR if you like.

docs/configuration.asciidoc
urso: s/stdErr/stderr/

Ah, got confused by formatting. No need to change it to RFC3339, but would be more consistent. If stderr and file logger would use same 'formatter'

ruflin: The above is the "standard" format provided for stdErr, so I would keep it as it is also very good readable. Should we use UTC or not here?

urso: I would include some timezone indicator. people get confused if time does not match their local time and no time zone info is present

ruflin: Lets make changes to the stderr logging in a separate pull request if required.

andrewkroh:
@ruflin (I will try to convince you of the syslog change.) I think that all logging is used to some degree for debugging and that is why I would like to have consistent logging irregardless of the destination. I feel that I shouldn't have to switch from syslog to file if I want to see files/line numbers in my DBG messages.

ruflin:
Ah, you would add it also to file logging, not only syslog. One good reason I see for this change is that it will simplify our logging and we have the same time of logs across all logs. Probably we should then also go with the same timestamp structure across all logs (see discussion above).

andrewkroh:
Yes, I think file and stderr outputs should use the same format for timestamp.

Issue #300, opened 2015-11-13
bug: direction field values are incorrect


rshriram:
The direction field indicates whether a transaction was incoming (recorded at the server) or outgoing (recorded the client). In its current form, the value of the direction field is incorrect. It shows "in" for outgoing transactions and "out" for incoming transactions.

I have sent a pull request with the respective bugfix.

andrewkroh:
Fixed in #301.

andrewkroh:
Change log is now updated.

PR #301, opened 2015-11-13
correctly determine direction of transaction

rshriram:
refer to bug #300 
This PR determines the direction of transaction correctly. Needs to be pushed to the rc1 branch.


andrewkroh:
@rshriram Thanks fixing this! I think we are going to make an RC2 next week so this will go into that.

PR #305, opened 2015-11-16
Improve error and log handling

ruflin:


CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ All notable changes to this project will be documented in this file based on the
 - Use of `enabled` as a configuration option for outputs (elasticsearch,
   logstash, etc.) has been removed. #264
 - Use of `disabled` as a configuration option for tls has been removed. #264
-- The `shipper` output field is renamed to `beat.name`. #285
```
andrewkroh: Why is this being removed here?

beat/beat.go
```diff
@@ -138,7 +140,7 @@ func (b *Beat) Run() {
 
 	service.Cleanup()
 
-	logp.Debug("beat", "Cleanup")
+	logp.Info("Cleaning up %s before shuttind down.", b.Name)
```
andrewkroh: `s/shuttind/shutting/`

common/geolite.go
```diff
@@ -15,21 +15,19 @@ type Geoip struct {
...
-	}
+	geoipPaths := []string{}
+
 	if config.Paths != nil {
-		geoip_paths = *config.Paths
+		geoipPaths = *config.Paths
 	}
-	if len(geoip_paths) == 0 {
+	if len(geoipPaths) == 0 {
 		// disabled
```
andrewkroh: I think this deserves an `logp.Info` message unless it is being logged elsewhere.

ruflin: No, it isn't logged somewhere else. Don't we normally log when something is enabled?

CHANGELOG.md
ruflin: It is still there (look at line 9). It was a duplicate.

beat/beat.go
ruflin: Fixed

common/geolite.go
ruflin: BTW: We should probably add a log message when it is enabled.

andrewkroh: Since it no longer automatically discovers the geoip database I was thinking it would be very helpful to log that no output::geoip::paths values were configured so there will be no geoip support.

I think there is already a message saying a file was loaded successfully.

urso: I think a Info is ok here. Info kinda logs what happens for you to know without really judging about the actual event.

common/geolite.go
```diff
@@ -15,21 +15,19 @@ type Geoip struct {
...
 	}
 
 	// look for the first existing path
-	var geoip_path string
-	for _, path := range geoip_paths {
+	var geoipPath string
+	for _, path := range geoipPaths {
 		fi, err := os.Lstat(path)
 		if err != nil {
 			continue
```
urso: maybe 'info'? It's an event about a file not readable or not found.

urso: the default value for geoip paths (if none if configured) consists of two pathes. Is it worth a warning (which is kinda error like already) if default values don't exist on your system?

urso: Nevermind. I see list of default paths is empty now.

PR #306, opened 2015-11-16
Remove beat.version

tsg:
After discussing it some more, we didn't have a clear use case for the version, and it's easier to add it in the future than remove it, so we prefer to do it before the GA.


urso:
+100

PR #314, opened 2015-11-18
Call shutdown handler only once

urso:
Fixes two ocassions where shutdown handler is called twice. If beats shutdown
callback did not already protect itself from multiple shutdown signals (e.g. when
closing channels) shutdown might result in panic.

- Seems like on windows the shutdown handler might be called either by signal
  handler or by windows service handler. Here go-routines are used => there
  might be a race between service shutdown handler and signal handler

- If second signal is received we don't want to send another signal to the beat
  (even if it is not as harmfull), to keep contract with beat and libbeat
  more simple

resolves elastic/filebeat#204

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file based on the
 ### Backward Compatibility Breaks
 
 ### Bugfixes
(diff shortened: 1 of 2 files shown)


urso:
PR for wrong branch.

PR #315, opened 2015-11-18
Optional CLIBeater interface

urso:
add optional CLIBeater interface. Using a cast we can figure out a beat wanting
to handle some CLI flags directly (e.g. packetbeat). If a Beater implements
CLIBeater the CLIFlags methods will called on beater right after parsing the
command line flags.

### DIFF
--- a/beat/beat.go
+++ b/beat/beat.go
@@ -22,6 +22,14 @@ type Beater interface {
 	Stop()
 }
 
+// Extended Beater interface for additional CLI handling before loading the
+// config file
+type CLIBeater interface {
+	Beater
+
+	CLIFlags(*Beat)
+}
+
 // Basic beat information
 type Beat struct {
 	Name    string
@@ -55,6 +63,30 @@ func NewBeat(name string, version string, bt Beater) *Beat {
 	return &b
 }
 
+// Initiates and runs a new beat object
+func Run(name string, version string, bt Beater) *Beat {
+	b := NewBeat(name, version, bt)
+
+	// Additional command line args are used to overwrite config options
+	b.CommandLineSetup()
+
+	// Loads base config
+	b.LoadConfig()
+
+	// Configures beat
+	err := bt.Config(b)
+	if err != nil {
+		logp.Critical("Config error: %v", err)
+		os.Exit(1)
+	}
+
+	// Run beat. This calls first beater.Setup,
+	// then beater.Run and beater.Cleanup in the end
+	b.Run()
+
+	return b
+}
+
 // Reads and parses the default command line params
 // To set additional cmd line args use the beat.CmdLine type before calling the function
 func (beat *Beat) CommandLineSetup() {
@@ -71,6 +103,11 @@ func (beat *Beat) CommandLineSetup() {
 	if *printVersion {
 		fmt.Printf("%s version %s (%s)\n", beat.Name, beat.Version, runtime.GOARCH)
 		os.Exit(0)
+	}
+
+	// if beater implements CLIFlags for additional CLI handling, call it now
+	if cliExtra, ok := beat.BT.(CLIBeater); ok {
+		cliExtra.CLIFlags(beat)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


urso:
Experimental change to libbeat Beats initialization.

Idea: The Beater interface defines the most minimal interface required by all beats. Additional setup/shutdown functionality (e.g. CLIFlags) can be added optionally to a beat by having additional interfaces. Using type casts libbeat dynamically finds these beats making use of these additional callbacks.

Having this change all beats (topbeat, filebeat, packetbeat) main.go becomes:

    package main

    import (
    	"github.com/elastic/libbeat/beat"

    	mybeat "github.com/elastic/mybeat/beat"
    )

    // You can overwrite these, e.g.: go build -ldflags "-X main.Version 1.0.0-beta3"
    var Version = "1.0.0"
    var Name = "mybeat"

    // Setups and Runs mybeat
    func main() {
    	beat.Run(Name, Version, mybeat.New())
    }

ruflin:
I like this change also without the CLIFlag interface. It makes creating a new beat even simpler. 

For the CLIFlag: Can you elaborate what would go into this interface? Are you referring to `flag.*`. Here be aware, that these must be defined in the init methods because of the system tests.

urso:
@ruflin this interface is used for packetbeat implementing some additional flag handling (e.g. for -devices)

ruflin:
If I get it right, this is the same as CommandLineSetup but beat specific? More thinking about the naming...

urso:
Hehe, naming is directly 'inspired' from packetbeat. [See offending line](https://github.com/elastic/packetbeat/blob/master/main.go#L26) in packetbeat its main.go

urso:
it's more like 'HandleFlags'

ruflin:
@urso what about this line? https://github.com/elastic/packetbeat/blob/master/main.go#L20

urso:
This line is not really required in main, so I moved into: packetbeat.New()

ruflin:
This change will make it much easier for new starters to create a beat. It will make it harder to understand what is going on for people that need more special implementations. But the second group will have a deeper understanding anyways.

My only issue is still about the CLIBeater naming. What you think about ExtendedBeater? The idea behind is, that flags will not be the only function that will be added here but for more complex beats additional functions will be needed. We could even look at the current Beater interface to look at which are not required to create a basic one. CLIBeater for me sounds like a beat that only runns on the command line.

ruflin:
@urso I suggest to merge this without the CLIBeater interface and discuss this based on my comment above in a second PR. Like this we can already start to migrate the beats to the new interface.

urso:
Oh, missed your comment about 'ExtendedBeater'. Nah, no fan of it. What's nice about interface is you can (and should) keep them at minimum. Maybe we can reduce the Beater interface in size too. 

e.g. CLIBeater can be changed to:

    type FlagsHandler interface {
        HandleFlags()
    }

This way we can add very fine grained 'extensions' a beat can optionally use. As I already did some POC of all beats migrating to this changeset, FlagsHandler will be required for packetbeat.

Only problem with this kind of change is documenting all potential interfaces a beat can implemented and how these interact on startup.

ruflin:
@urso I like the multiple small interfaces approach. Lets go with FlagsHandler. Seems to me more intuitive. We should be able to cover it in the developer guide.

andrewkroh:
I also like the `FlagsHandler` naming.

urso:
rename CLIBeater to FlagsHandler. Ready for merging.

This changeset is fully BC (well, should be).

beat/beat.go
```diff
@@ -22,6 +22,14 @@ type Beater interface {
 	Stop()
 }
 
+// FlagsHandler (optional) Beater extension for
+// handling flags input on startup. The HandleFlags callback will
+// be called after parsing the command line arguments and handling
```
andrewkroh: Does this mean that the callback will be invoked after `-help` or `-version` are handled? It reads like that to me, but I think the beat actually exits after processing help and version.

urso: yes, callback will be handled after '-help' and '-version' are handled, if set.

PR #322, opened 2015-11-23
Fix wrong types in debug log messages

ruflin:


outputs/logstash/client.go
```diff
@@ -76,7 +76,7 @@ func (l *lumberjackClient) PublishEvents(
 	for len(events) > 0 {
 		n, err := l.publishWindowed(events)
 
-		logp.Debug("logstash", "%s events out of %s events sent to logstash. Continue sending ...", n, len(events))
+		logp.Debug("logstash", "%d events out of %d events sent to logstash. Continue sending ...", n, len(events))
```
urso: why not use '%v'?

ruflin: changed to %v

PR #323, opened 2015-11-24
Fix elasticsearch ping authentication fail

urso:
Have ping use execRequest, so credentials are correctly set in request. Also
add some debug logging, for identifying reasons connection fail.

resolves elastic/filebeat#287


urso:
wrong branch

PR #328, opened 2015-11-24
Clarifies guidance about when to update Logstash Input Beats plugin

dedemorton:
Also updated example response from Elastic because the version info in the example was old. Added a few other minor fixes, too


monicasarbu:
great, thanks @dedemorton

PR #340, opened 2015-12-01
Fix typo

gabriel-tessier:
/!\ Rpm and mac package are with "-" only .deb file have a "_" !!


ruflin:
@gabriel-tessier Thanks for spotting and fixing this.

PR #341, opened 2015-12-01
Set a default config file location for Windows

andrewkroh:
Fixes elastic/winlogbeat#2.

On Windows you will get this:
```
PS C:\Gopath\src\github.com\elastic\winlogbeat> .\build\bin\winlogbeat-windows-386.exe -h
Usage of C:\Gopath\src\github.com\elastic\winlogbeat\build\bin\winlogbeat-windows-386.exe:
  -N    Disable actual publishing for testing
  -c string
        Configuration file (default "C:\\Program Files\\Winlogbeat\\winlogbeat.yml")
  -configtest
        Test configuration and exit.
  -cpuprofile string
        Write cpu profile to file
  -d string
        Enable certain debug selectors
  -e    Log to stderr and disable syslog/file output
  -memprofile string
        Write memory profile to this file
  -v    Log at INFO level
  -version
        Print version and exit
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file based on the
 ### Backward Compatibility Breaks
 
 ### Bugfixes
+- Fix default config file path for Windows. #341
 
 ### Added
 
--- a/cfgfile/cfgfile.go
+++ b/cfgfile/cfgfile.go
@@ -4,6 +4,8 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"runtime"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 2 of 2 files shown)


ruflin:
It would be nice to handle this somehow in beats-packer instead of directly in the code. The basic "assumption" I would think is that the file is in the same directory as the binary, as this is what happens after the download. Could we somehow inject this path during compiling?

andrewkroh:
I think the behavior from this PR is consistent with linux. If we change the default for Windows to be the directory in with the binary resides I would do the same across all operating systems, but I like the behavior as is.

The Windows installer (PowerShell script at this time) does take into account the fact that the config file is delivered in the same directory as the binary and automatically sets up the Windows service accordingly.

ruflin:
@andrewkroh I would also remove the Linux part and have a system independent default. As far as I know, elasticsearch assumes after the download that it is in the config folder. I tried to find the exact code, but didn't fit it yet. Here is something related: https://github.com/elastic/elasticsearch/blob/5337686d8d39568f94b0b976bd9098af69214302/core/src/main/java/org/elasticsearch/bootstrap/Bootstrap.java#L360

So in case it is installed with the deb package and a startup script, the startup script would handle the "dynamic" path by using -c.

andrewkroh:
@ruflin What you describe makes sense to make. In summary,

- Set the `-c` flag in the startup or service scripts to customize the config path on an OS install basis.
- Make the binary look for a config file in the same directory as the binary.

ruflin:
@andrewkroh exactly. I opened an issue here for it: https://github.com/elastic/beats/issues/419

PR #342, opened 2015-12-02
Adding the ability to use a proxy server when connecting to ES

### LINKED ISSUE #338: Proxy Server Configuration
There does not appear to be a way to specify a HTTP proxy server to use when connecting to Elasticsearch.

As a workaround I think the `HTTP_PROXY` and `HTTPS_PROXY` environment variables can be set, but it would be better if the proxy server could be set through the configuration file. On Windows, it is particularly important to have them in the configuration file because setting the env vars for services is a PITA because it requires a system restart.

andrewkroh:
Fixed #338

If you don't want to merge this prior to reorganizing repos, I can easily open it up against the new repo. I tested it locally against a squid server, but it would be nice to add an integration test for it.

### DIFF
--- a/outputs/elasticsearch/api_mock_test.go
+++ b/outputs/elasticsearch/api_mock_test.go
@@ -48,7 +48,7 @@ func TestOneHostSuccessResp(t *testing.T) {
 
 	server := ElasticsearchMock(200, expectedResp)
 
-	client := NewClient(server.URL, "", nil, "", "")
+	client := NewClient(server.URL, "", nil, nil, "", "")
 
 	params := map[string]string{
 		"refresh": "true",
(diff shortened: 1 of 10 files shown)


urso:
what if the proxy requires authentication?

outputs/elasticsearch/api_mock_test.go
```diff
@@ -48,7 +48,7 @@ func TestOneHostSuccessResp(t *testing.T) {
 
 	server := ElasticsearchMock(200, expectedResp)
 
-	client := NewClient(server.URL, "", nil, "", "")
+	client := NewClient(server.URL, "", nil, nil, "", "")
```
urso: a pity unit tests need to be updated. Maybe have a newTestClient() with default. 

6 parameters is quite a lot. At some point we should introduce a struct for the client configuration.

andrewkroh:
Regarding authentication, that is a good question that does not seem to be addressed in the golang documentation. Lets keep #338 open and make sure we address proxy authentication before closing. For basic auth I am hoping you can just add it to the proxy URL like `http://user:pass@proxy:3128`.

ruflin:
LGTM

PR #345, opened 2015-12-03
Support customizing beat directory used in 'make update'

mrkschan:


ruflin:
@mrkschan We merged all beats repo into one. The files now moved here: https://github.com/elastic/beats/tree/master/libbeat

Also it is probably good to wait for this PR to be merged: https://github.com/elastic/beats/pull/434

It would be still very nice if the same script can be used for your beats.

mrkschan:
Eh, so my godep should points to github.com/elastic/beats/libbeat/..., right?

ruflin:
@mrkschan yes. Let me know if other things break.

